PR #13, opened 2020-01-17
Fixes #10 -- implement task routing

### LINKED ISSUE #10: Set up a task router
While implementing #8, we receive back tasks from different topics. While we can loop over all known topics and fetch and lock tasks like that, it's probably more robust to set up a task router based on topic that knows where to send tasks for further processing.

Hooking that into celery tasks/workers will then provide sufficient scaling options.

sergei-maertens:


src/camunda_worker/tasks/api.py
```diff
@@ -0,0 +1,60 @@
...
+
+    try:
+        handler = registry[task_mapping.callback]
+    except KeyError as exc:
+        raise NoCallback(
+            f"Callback '{task_mapping.callback}' is not in the provided registry."
+        ) from exc
+
+    # check for expiry
+    if task.expired:
```
annashamray: Should the check for `task.status` be also made here?

annashamray: And perhaps changing `task.status` and saving `task.result_variables` should be on this level?

sergei-maertens: I don't think so, if you're retrying a task because it failed before, that shouldn't block it? :thinking:

annashamray: But in this case the status would be `failed`, right? I think it'd be nice to have the task state logic somewhere. Or remove `task.status` field if it is not used

PR #22, opened 2020-01-31
initial config for celery

annashamray:


bin/celery_worker.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/sh
```
sergei-maertens: `docker_start.sh` still uses sh too, but we can actually use `bash` instead of `sh` since Debian is used as base image :)

bin/celery_worker.sh
```diff
@@ -0,0 +1,16 @@
...
+LOGLEVEL=${CELERY_LOGLEVEL:-INFO}
+
+mkdir -p ../celerybeat
+
+echo "Starting celery worker"
+celery worker \
+    --app camunda_worker \
+    -l $LOGLEVEL \
+    --workdir src \
+    -B \
```
sergei-maertens: Can you create a separate `celery_beat.sh` script? See https://docs.celeryproject.org/en/latest/reference/celery.bin.worker.html#cmdoption-celery-worker-b

annashamray:
@sergei-maertens thank you for the check-list! It helped a lot

> Looks good for initial setup, couple things are missing:
> 
> 1. Can you add the shell scripts to the Docker image by modifying `Dockerfile`
> 2. Make sure that the `celery` binary is also available in the final docker image
> 3. Update the docker-compose.yml to also run the celery worker and beat :-)

Dockerfile
```diff
@@ -58,12 +58,14 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
 
 WORKDIR /app
 COPY ./bin/docker_start.sh /start.sh
+COPY ./bin/celery_worker.sh /celery_worker.sh
```
sergei-maertens: you're missing the `celery_beat` script still :sweat_smile:

docker-compose.yml
```diff
@@ -33,9 +33,28 @@ services:
...
-#    ports:
-#        - 6379:6379
+
+  redis:
+    # NOTE: No persistance storage configured.
+    image: redis
+
+  celery:
+    build: .
+    environment: &celery_env
```
sergei-maertens: you'll probably want to register that as `&web_env` and re-use that - since the web should also know which broker is used etc.

annashamray: done

PR #24, opened 2020-02-05
Feature/set celery tasks

annashamray:


bin/celery_worker.sh
```diff
@@ -9,3 +9,4 @@ celery worker \
     --app camunda_worker \
     -l $LOGLEVEL \
     --workdir src \
+    -B
```
sergei-maertens: hrm, this should only be for dev purposes and not run in production, right?

src/camunda_worker/tasks/migrations/0003_change_callback_path.py
```diff
@@ -0,0 +1,25 @@
...
+        mapping.callback = ".".join(bits)
+        mapping.save()
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [("tasks", "0002_auto_20200120_1240")]
+
+    operations = [
+        migrations.RunPython(rewrite_callback_paths, migrations.RunPython.noop)
```
sergei-maertens: can you correctly implement the reverse migration insead of `noop`, or leave it out so that migrations can not be reverted?

I could now migrate, reverse, and migrate again and then I get `.worker.worker` in the callback.

src/camunda_worker/tasks/tasks/celery.py
```diff
@@ -0,0 +1,59 @@
...
+
+
+@app.task()
+def task_fetch_and_lock():
+    worker_id, num_tasks, tasks = fetch_and_lock(settings.MAX_TASKS)
+
+    logger.info("fetched %r tasks with %r", num_tasks, worker_id)
+
+    for task in tasks:
+        task_execute_and_complete(task.id)
```
sergei-maertens:
```suggestion
        task_execute_and_complete.delay(task.id)
```

I think the fetching and locking should be decoupled from the actual execution - now the execution is done synchronously as part of the fetch and lock task.

src/camunda_worker/tasks/tasks/celery.py
```diff
@@ -0,0 +1,59 @@
...
+    return num_tasks
+
+
+@app.task()
+def task_execute_and_complete(fetched_task_id):
+    fetched_task = FetchedTask.objects.get(id=fetched_task_id)
```
sergei-maertens: I'd filter this queryset down to tasks that have a status that qualifies them for processing - so you don't accidentally execute a completed task.

src/camunda_worker/tasks/tasks/celery.py
```diff
@@ -0,0 +1,59 @@
...
+
+    fetched_task.status = Statuses.in_progress
+    fetched_task.save()
+
+    # execute
+    try:
+        execute(fetched_task)
+    except Exception as exc:
+        logger.debug("Task %r has failed with error: %r", fetched_task_id, exc)
```
sergei-maertens:
```suggestion
        logger.debug("Task %r has failed with error: %r", fetched_task_id, exc, exc_info=True)
```
will include the exception details in the logging setup as well.

src/camunda_worker/tasks/tasks/celery.py
```diff
@@ -0,0 +1,59 @@
...
+
+        return
+
+    logger.info("Task %r is executed", fetched_task_id)
+
+    # complete
+    try:
+        complete(fetched_task)
+    except Exception as exc:
+        logger.debug("Task %r has failed with error: %r", fetched_task_id, exc)
```
sergei-maertens: I'd avoid using the exact same logging message - `execute` and `complete` had very different natures

src/camunda_worker/tasks/tasks/celery.py
```diff
@@ -0,0 +1,59 @@
...
+
+        fetched_task.status = Statuses.failed
+        fetched_task.save()
+
+        return
+
+    logger.info("Task %r is completed", fetched_task_id)
```
sergei-maertens: shouldn't you also set the `FetchedTask` status to completed?

annashamray: statuses `performed` and `completed` are set in the `tasks/api.py`

src/camunda_worker/tasks/tasks/celery.py
annashamray: added more details to the description of failed tasks

src/camunda_worker/tasks/tasks/celery.py
annashamray: the task invokes `execute` and `complete` api, which process different statuses and can raise specified exceptions if the statuses are not correct

src/camunda_worker/tasks/migrations/0003_change_callback_path.py
annashamray: `reverse_code` is removed from operation

src/camunda_worker/tasks/tasks/celery.py
sergei-maertens: right, forgot about that :-)

PR #25, opened 2020-02-10
Big rename

annashamray:


annashamray:
> Also, impressive that the code coverage went up with a rename ;)

@joeribekker yes! I don't know it works.
The README file is updated

PR #32, opened 2020-02-20
Fetch & lock fixes / ensure result variables are set on task complete

sergei-maertens:
* Tweaked fetch and lock so that only configured topics are fetched & locked
* Make sure to pass the `result_variables` to the complete call
* Modified execute_task(s) management command, basically changed completely. We have celery beat & `fetch_and_lock_tasks` management commands to grab them, this makes debugging a bit easier.


src/bptl/camunda/management/commands/execute_task.py
```diff
@@ -0,0 +1,24 @@
...
+    def handle(self, **options):
+        task_id = options["task_id"]
+        task = ExternalTask.objects.filter(id=task_id).first()
+        if task is None:
+            raise CommandError("Could not find this task.")
+
+        try:
+            task_execute_and_complete(task_id)
+        except Exception as exc:
+            raise CommandError(exc) from exc
```
annashamray: the exception related to execution won't be raise because it would be caught by `task_execute_and_complete` celery task. I think this `try except` is unnecessary.

annashamray: Can you add some logging of what is happening during  the command execution? 
In both cases if the task is complete or if it failed there is no feedback in the stdout. 

The celery task logs the result but with celery logger, which is not used here.

PR #33, opened 2020-02-24
Feature/data for monitoring

annashamray:


src/bptl/camunda/tests/test_celery_tasks.py
```diff
@@ -39,6 +41,7 @@ def test_task_execute_and_complete_success(self, m_execute, m_complete):
     @patch("bptl.camunda.tasks.complete")
     @patch("bptl.camunda.tasks.execute", side_effect=Exception("execution is failed"))
     def test_task_execute_and_complete_fail_execute(self, m_execute, m_complete):
+        bptl.camunda.tasks.execute = save_and_log()(bptl.camunda.tasks.execute)
```
sergei-maertens: This seems like a bad idea - you're effectively replacing what's defined in the module but on a global scale. Pretty sure this breaks test isolation.

Probably better to do something like:

```suggestion
    @patch("bptl.camunda.tasks.complete")
    @patch("bptl.camunda.tasks.execute")
    def test_task_execute_and_complete_fail_execute(self, m_execute, m_complete):

        @save_and_log()
        def new_execute(task):
            raise Exception("execution is failed")

        m_execute.side_effect = new_execute
```

src/bptl/camunda/tests/test_celery_tasks.py
```diff
@@ -53,6 +56,7 @@ def test_task_execute_and_complete_fail_execute(self, m_execute, m_complete):
     @patch("bptl.camunda.tasks.complete", side_effect=Exception("sending is failed"))
     @patch("bptl.camunda.tasks.execute")
     def test_task_execute_and_complete_fail_complete(self, m_execute, m_complete):
+        bptl.camunda.tasks.complete = save_and_log()(bptl.camunda.tasks.complete)
```
sergei-maertens: see above

PR #34, opened 2020-02-25
Feature/dashboard api

annashamray:
depends on https://github.com/GemeenteUtrecht/bptl/pull/33


src/bptl/dashboard/views.py
```diff
@@ -0,0 +1,46 @@
...
+        item_data = {q["status"]: q["tasks"] for q in item_qs}
+        items.append({type: item_data})
+
+    total_qs = BaseTask.objects.annotate_status()
+    total_data = {q["status"]: q["tasks"] for q in total_qs}
+
+    return {"items": items, "total": total_data}
+
+
+class AggregateView(APIView):
```
sergei-maertens: I don't like mixing API endpoints and non-API views/urls. Can you put the API stuff in an `api/` package?

src/bptl/urls.py
```diff
@@ -37,7 +37,8 @@
     ),
     # Simply show the master template.
     path("", TemplateView.as_view(template_name="index.html"), name="index"),
-    path("tasks/", include("bptl.tasks.urls")),
+    path("tasks/", include("bptl.dashboard.urls")),
```
sergei-maertens: Please separate API endpoints from regular HTML views/URLs :)

src/bptl/dashboard/views.py
annashamray: `AggregateView` is moved to `dashboard/api` package

src/bptl/urls.py
annashamray: done

PR #38, opened 2020-02-28
create pie chart in dashboard

annashamray:


src/bptl/js/components/pie-chart.js
```diff
@@ -0,0 +1,150 @@
...
+        d3.select("#piechart")
+          .selectAll("*")
+          .remove();
+
+        const svg = d3.select("#piechart")
+          .append("svg")
+            .attr("width", width)
+            .attr("height", height)
+          .append("g")
+            .attr("transform", "translate(" + width / 2 + "," + height / 2 + ")");
```
sergei-maertens: FYI - Javascript also knows template literals, e.g.:

```javascript
const translation = `translate(${width / 2},${height / 2})`;
```

PR #39, opened 2020-02-28
Task tweaking

sergei-maertens:


src/bptl/conf/base.py
```diff
@@ -369,7 +370,7 @@
 CELERY_BEAT_SCHEDULE = {
     "task-pull": {
         "task": "bptl.camunda.tasks.task_fetch_and_lock",
-        "schedule": crontab(minute="*"),
+        "schedule": schedule(run_every=1),  # run every second
```
sergei-maertens: This will be upped again later when we implement long-polling.

sergei-maertens:
Created #40 for this

PR #43, opened 2020-03-09
Feature/add default services to form

annashamray:


src/bptl/tasks/tests/test_add_defaultservices.py
```diff
@@ -0,0 +1,142 @@
...
+    def setUp(self):
+        self.client.login(username="super", password="letmein")
+
+        patcher = patch("bptl.tasks.forms.register", new=test_register)
+        self.mocked_register = patcher.start()
+        self.addCleanup(patcher.stop)
+
+        self.url = reverse_lazy("tasks:taskmapping-create")
+
+        self.data = {
```
sergei-maertens: For the future: django-webtest makes your life so much easier when dealing with forms/formsets :)

src/bptl/tasks/views.py
```diff
@@ -19,3 +19,29 @@ class AddTaskMappingView(LoginRequiredMixin, CreateView):
     model = TaskMapping
     form_class = TaskMappingForm
     success_url = reverse_lazy("tasks:taskmapping-list")
+
+    def get_formset(self):
```
sergei-maertens: django-extra-views has some useful base views to deal with formsets, please look at that next time :)

src/bptl/tasks/tests/test_add_defaultservices.py
annashamray: I know, but when I tried to use `django-webtest` in response.form the `service` fields were missing (`alias` fields were there though). In browser everything was ok, so I just switched to django testcases

src/bptl/tasks/views.py
annashamray: Thanks! I'll tasks a look

PR #45, opened 2020-03-11
add IsAboveAge task

annashamray:


src/bptl/work_units/brp/tasks.py
```diff
@@ -0,0 +1,39 @@
...
+
+from .client import get_client_class
+
+__all__ = ["IsAboveAge"]
+
+
+@register
+class IsAboveAge(WorkUnit):
+    """
+    Fetches BRP API and return flag if a certain person is older above a certain age
```
joeribekker: This help text gets picked up by the UI right? In that case, I think we should also mention here it's "equal or above".

```suggestion
    Fetches BRP API and returns whether a person is exactly, or older than, a certain age.
```

src/bptl/work_units/zgw/tasks.py
```diff
@@ -264,6 +264,7 @@ class RelatePand(ZGWWorkUnit):
     **Required process variables**
 
     * ``zaakUrl``: URL reference to a ZAAK in a Zaken API. The PANDen are related to this.
+    * ``panden``: list of URL references to PANDen in BAG API.
```
joeribekker: Left over from elsewhere?

annashamray: It is a task from Utrecht hackathon. I just spotted that it lacks a description for one of variables

Issue #58, opened 2020-05-18
Rework the on-behalf-of credentials system


sergei-maertens:
Current issues:

- tokens are included in the process variables, where they can be read by anyone with access to the process instance
- tokens expire after an hour (for ZGW at least), so if the process crashes and is retried past that, the tokens need to be updated with valid ones
- each work unit needs to be connected with the relevant services _and_ aliases need to be set. These aliases need to be communicated out of band to the applications setting the variables. This is bound to go wrong

Current direction of solution:

1. Application *currently* handling the process sets a variable: `getCredentialsUrl`.
    This should be a URL that is impossible to guess, and ideally be based around a hash of some attribute that changes when accessed/when credentials have been obtained -> we want to prevent replay attacks. Possibly BPTL itself needs to be authorized to the application URL out of band with a static token.
2. Whenever BPTL needs credentials, it makes a POST request:
    
    ```http
    POST /${getCredentialsUrl} HTTP/1.1

    {"api": "https://example.com/api/v1/"}
    ```

    The application replies with the relevant headers required, e.g.:

    ```json
    {"header": "Authorization", "value": "Bearer <some-jwt>", "api": "https://example.com/api/v1/"}
    ```
3. BPTL uses those credentials to make the required API calls
4. Whenever other applications take over for the process, they can set their own `getCredentialsUrl` process variable to indicate that their credentials should now be used (relevant for audit logs!).

This solves:
- no need to use aliases
- having the URL as process variable is not enough for a replay attack
- "refreshing" credentials becomes straight-forward, as they are retrieved when they are needed (and generated on the server if relevant)
- each unit knows which services it needs. BPTL can still configure which particular service to use for a topic if there are multiple services providing the same functionality but with different storage.

PR #61, opened 2020-07-13
Valid sign integration

SilviaAmAm:


src/bptl/work_units/valid_sign/tasks.py
```diff
@@ -0,0 +1,264 @@
...
+
+    def create_package(self) -> dict:
+        """
+        Creates a ValidSign package with the name specified by the process variables and adds the signers to it.
+        """
+        logger.debug("Creating ValidSign package")
+
+        variables = self.task.get_variables()
+        signers = self.format_signers(variables.get("signers"))
+        package_name = variables.get("package_name")
```
sergei-maertens, line 122: if the variable has to be present, use `check_variable` instead: see https://business-process-task-library.readthedocs.io/en/latest/work_units.html#variable-extraction

src/bptl/work_units/valid_sign/tasks.py
```diff
@@ -0,0 +1,264 @@
...
+
+        url = f"{settings.VALIDSIGN_ROOT_URL}api/packages"
```
sergei-maertens, line 124: should not use settings

src/bptl/work_units/valid_sign/tasks.py
```diff
@@ -0,0 +1,264 @@
...
+
+        body = {
+            "name": package_name,
+            "type": "PACKAGE",
+            "roles": signers,
+        }
+
+        response = requests.post(url, headers=self._auth_header, data=json.dumps(body))
```
sergei-maertens, line 132: zgw consumers client abstracts this away, but even with barebones requests you should do:

```suggestion
        response = requests.post(url, headers=self._auth_header, json=body)
```

src/bptl/work_units/valid_sign/tasks.py
```diff
@@ -0,0 +1,264 @@
...
+        response.raise_for_status()
+        package = response.json()
+
+        return package
+
+    def add_documents_to_package(self, package: dict) -> List[dict]:
+        """
+        Adds documents to the specified package and returns a list with the information about each document returned by
```
sergei-maertens, line 140: Docstring PEP257

src/bptl/work_units/valid_sign/tasks.py
```diff
@@ -0,0 +1,264 @@
...
+        ValidSign.
+        """
+
+        logger.debug(f"Adding documents to ValidSign package '{package.get('id')}'")
```
sergei-maertens, line 144: Logging call/interpolation issue as above

src/bptl/work_units/valid_sign/tasks.py
```diff
@@ -0,0 +1,264 @@
...
+
+        documents = self._get_documents_from_api()
+
+        post_url = (
+            f"{settings.VALIDSIGN_ROOT_URL}api/packages/{package.get('id')}/documents"
```
sergei-maertens, line 149: should not use settings

src/bptl/work_units/valid_sign/tasks.py
```diff
@@ -0,0 +1,264 @@
...
+        # to the request, but then not sure how to specify the filename yet...
+        # files = [("files[]", content) for name, content in documents]
+
+        attached_documents = []
+        for doc_name, doc_content in documents:
+            payload = {"payload": f'{{"name": "{doc_name}"}}'}
+            file = [("file", doc_content)]
+            response = requests.post(
+                post_url, headers=self._auth_header, files=file, data=payload
+            )
```
sergei-maertens, line 162: I don't trust the input here - a `doc_name` with value ` fo"o` will break the JSON (add a test for this!)

```suggestion
            payload = {"name": doc_name}
            post_data = {"payload": json.dumps(payload)}
            file = [("file", doc_content)]
            response = requests.post(
                post_url, headers=self._auth_header, files=file, data=post_data
            )
```

src/bptl/work_units/valid_sign/tasks.py
```diff
@@ -0,0 +1,264 @@
...
+            response.raise_for_status()
+            attached_documents.append(response.json())
+
+        return attached_documents
+
+    def create_approval_for_documents(self, package: dict, documents: List[dict]):
+        """
+        Creates an approval (a placeholder for where a signature will go) in the
```
sergei-maertens, line 170: PEP257 :)

src/bptl/work_units/valid_sign/tasks.py
```diff
@@ -0,0 +1,264 @@
...
+        specified documents for all signers.
+
+        According to https://apidocs.validsign.nl/validsign_integrator_guide.pdf the anchor extraction cannot be used
+        with the API call to create an approval. So, the position has to be provided. If no ``top``, ``left``,
+        ``width`` and ``height`` are given, then an 'acceptance button' appears under the document.
+        """
+
+        logger.debug(
+            f"Creating approvals for documents in ValidSign package '{package.get('id')}'"
```
sergei-maertens, line 179: Log-call format, see earlier

src/bptl/work_units/valid_sign/tasks.py
```diff
@@ -0,0 +1,264 @@
...
+        signers = self._get_signers_from_package(package)
+
+        # Settings for the size and the place of the signature field in the document
+        signature_width = 150
+        signature_height = 50
+        left_offset = 0
+
+        # For all the documents, create an approval for each signer
+        for document in documents:
+            approval_url = f"{settings.VALIDSIGN_ROOT_URL}api/packages/{package.get('id')}/documents/{document.get('id')}/approvals"
```
sergei-maertens, line 191: - shouldn't use settings
- shouldn't use the `.get` form

src/bptl/work_units/valid_sign/tasks.py
```diff
@@ -0,0 +1,264 @@
...
+                        "top": counter * signature_height,
+                        "left": left_offset,
+                        "width": signature_width,
+                        "height": signature_height,
+                        "page": 0,
+                        "type": "SIGNATURE",
+                        "subtype": "FULLNAME",
+                    }
+                ]
+                data = {"role": f"{signer.get('id')}", "fields": approval_settings}
```
sergei-maertens, line 205:
```suggestion
                data = {"role": f"{signer['id']}", "fields": approval_settings}
```

src/bptl/work_units/valid_sign/tasks.py
```diff
@@ -0,0 +1,264 @@
...
+                response = requests.post(
+                    approval_url, data=json.dumps(data), headers=self._auth_header
```
sergei-maertens, line 207:
```suggestion
                    approval_url, json=data, headers=self._auth_header
```

src/bptl/work_units/valid_sign/tasks.py
```diff
@@ -0,0 +1,264 @@
...
+                )
+                response.raise_for_status()
+
+    def send_package(self, package: dict):
+        """
+        Changes the status of a package to "SENT". This automatically sends an email to all the signers with a
```
sergei-maertens, line 213: PEP 257

src/bptl/work_units/valid_sign/tasks.py
```diff
@@ -0,0 +1,264 @@
...
+        link where they can sign the documents.
+        """
+        logger.debug(f"Setting the status of package '{package.get('id')}' to SENT")
```
sergei-maertens, line 216: logger formatting, `.get` call

src/bptl/work_units/valid_sign/tasks.py
```diff
@@ -0,0 +1,264 @@
...
+        url = f"{settings.VALIDSIGN_ROOT_URL}api/packages/{package.get('id')}"
```
sergei-maertens, line 217: shouldn't use settings but a zgw-consumers client

src/bptl/work_units/valid_sign/tasks.py
```diff
@@ -0,0 +1,264 @@
...
+        body = {"status": "SENT"}
+        response = requests.put(url, headers=self._auth_header, data=json.dumps(body))
```
sergei-maertens, line 219:
```suggestion
        response = requests.put(url, headers=self._auth_header, json=body)
```

src/bptl/work_units/valid_sign/tasks.py
```diff
@@ -0,0 +1,264 @@
...
+        response.raise_for_status()
+
+    def perform(self) -> dict:
+        package = self.create_package()
+        documents = self.add_documents_to_package(package)
+        self.create_approval_for_documents(package, documents)
+        self.send_package(package)
+
+        return {"package_id": package.get("id")}
```
sergei-maertens, line 228:
```suggestion
        return {"packageId": package["id"]}
```

src/bptl/work_units/valid_sign/tasks.py
```diff
@@ -0,0 +1,264 @@
...
+
+
+class ValidSignReminderTask(WorkUnit):
+    """
+    Takes a ValidSign package ID and the email address of a signer and sends an email-reminder to the signer providing a
+    link to where the documents can be signed.
+
+    **Required process variables**
+
+    * ``package_id``: string with the ValidSign Id of a package
```
sergei-maertens, line 238:
```suggestion
    * ``packageId``: string with the ValidSign Id of a package
```

src/bptl/work_units/valid_sign/tasks.py
```diff
@@ -0,0 +1,283 @@
...
+    _documents_client = None
+
+    def get_documents_client(self, topic_name: str) -> ZGWClient:
+        if self._documents_client is None:
+            default_services = TaskMapping.objects.get(
+                topic_name=topic_name
+            ).defaultservice_set.select_related("service")
+            services_by_alias = {svc.alias: svc.service for svc in default_services}
+
+            alias = "DocumentenAPI"
```
sergei-maertens, line 79: We typically use `drc` for this alias.

```suggestion
            alias = "drc"
```

src/bptl/work_units/valid_sign/tasks.py
```diff
@@ -0,0 +1,283 @@
...
+        return [{"type": "SIGNER", "signers": [signer]} for signer in signers]
+
+    def _get_documents_from_api(self) -> List[Tuple[str, bytes]]:
+        """Retrieve the documents and their content from the Documenten API."""
+
+        logger.debug("Retrieving documents from Documenten API")
+
+        variables = self.task.get_variables()
+        document_urls = check_variable(variables, "documents")
+        document_client = self.get_documents_client(topic_name="CreateValidSignPackage")
```
sergei-maertens, line 98: hard-coded topic name is not going to work - the topic name is dynamic and specified when you create `TaskMapping` objects. You can retrieve it from `self.task` though.

src/bptl/work_units/valid_sign/tasks.py
```diff
@@ -0,0 +1,283 @@
...
+
+        documents = []
+        for document_url in document_urls:
+            # Retrieving the document
+            document_data = document_client.retrieve(
+                resource="enkelvoudiginformatieobject", url=document_url,
+            )
```
sergei-maertens, line 105: This is going to be a rough one - it's not unrealistic that multiple Documents API's are used, for example:
- https://drc.utrechtproeftuin.nl/api/v1/
- https://alfresco.utrechtproeftuin.nl/drc/api/v1/
- https://openzaak.utrechtproeftuin.nl/documenten/api/v1/

Each of those have different credentials. The client should be constructed from `document_url` rather than using a static configured client.

It's a complex situation, especially w/r to #58 in mind - the JWT's have to be fed into the process from the initial applications.

What you _can_ assume is:

- each DRC service is connected to the task mapping (`TaskMapping` and `DefaultService`) with an alias, e.g. `drc1`, `drc2`...
- the process variable `services` contains the JWT's - see how the `zgw` work units do this

Then, you can:

1. for every default service for the task mapping
    1. build the client
2. find the best match within the default services (check if URL starts with `client.api_root`)
3. use that client and use the credentials from the `services` variable

This can be abstracted away properly in zgw-consumers, looking at https://github.com/maykinmedia/zgw-consumers/blob/master/zgw_consumers/query.py#L10 by turning this manager method into a queryset method and filtering the services based on the task default services.

src/bptl/work_units/valid_sign/tasks.py
```diff
@@ -0,0 +1,283 @@
...
+
+            # Retrieving the content of the document
+            # Need use requests directly instead of `document_client.request()` since the response is not in JSON format
+            response = requests.get(
+                document_data["inhoud"], headers=document_client.auth.credentials(),
+            )
+
+            documents.append((document_data["titel"], response.content))
```
sergei-maertens, line 113: I am curious what will happen if there are many large files being kept in memory - that's probably going to OOM-kill the container.

We should probably have a treshold of maximum filesize, and above that file size write a temporary file (see https://docs.djangoproject.com/en/3.0/ref/files/uploads/#django.core.files.uploadedfile.TemporaryUploadedFile) and read the file again when we send to ValidSign.

Additionally, there should probably be a global limit (like if you have 1000 docs each 1 MB, that's still 1GB in memory!).

src/bptl/work_units/valid_sign/tasks.py
```diff
@@ -0,0 +1,283 @@
...
+
+        return documents
+
+    def _get_signers_from_package(self, package: dict) -> List[dict]:
+        """Retrieve all the roles from a ValidSign package and return those that are signers."""
+
+        logger.debug("Retrieving the roles from validSign package '%s'", package["id"])
+
+        response = self.get_validsign_client(
+            topic_name="CreateValidSignPackage"
```
sergei-maertens, line 123: topic name should be derive from `self.task`

src/bptl/work_units/valid_sign/tasks.py
```diff
@@ -0,0 +1,283 @@
...
+        )
+
+        documents = self._get_documents_from_api()
+
+        # Multiple files can be added in one request by passing the following 'files' parameter
+        # to the request, but then not sure how to specify the filename yet...
+        # files = [("files[]", content) for name, content in documents]
+
+        validsign_client = self.get_validsign_client(
+            topic_name="CreateValidSignPackage"
```
sergei-maertens, line 200: hardcoded topic name is not ok :)

src/bptl/work_units/valid_sign/tasks.py
SilviaAmAm: Updated!

src/bptl/work_units/valid_sign/tasks.py
SilviaAmAm: Updated!

src/bptl/work_units/valid_sign/tasks.py
SilviaAmAm: Updated

src/bptl/work_units/valid_sign/tasks.py
SilviaAmAm: Updated :)

src/bptl/conf/base.py
```diff
@@ -418,3 +419,9 @@
         "Token": {"type": "apiKey", "name": "Authorization", "in": "header"}
     },
 }
+
+# ValidSign Documents settings
+# The maximum size (in Mb) that all the documents combined can have
+MAX_TOT_DOCUMENT_SIZE = os.getenv("MAX_TOT_DOCUMENT_SIZE", 2000)
```
sergei-maertens, line 425: 2000 Mb is _a lot_ - that requires that your process is allowed to use 2GB of memory. I'd change that to about 250 Mb instead.

src/bptl/work_units/valid_sign/tasks.py
```diff
@@ -0,0 +1,382 @@
...
+            # the maximum size, write the file content to a temporary file
+            if (
+                document_size > settings.MAX_DOCUMENT_SIZE
+                or (current_total_documents_size + document_size)
+                > settings.MAX_TOT_DOCUMENT_SIZE
+            ):
+                tmp_file_object = TemporaryUploadedFile(
+                    name=f"{document_data['titel']}-{get_random_string(length=5)}.tempfile",
+                    content_type="application/octet-stream",
+                    size=sys.getsizeof(response.content),
```
sergei-maertens, line 197: accessing `response.content` causes the entire document to be loaded into memory and is precisely what we don't want to do to prevent out-of-memory errors.

`document_data['bestandsomvang']` has the filesize in bytes, per the API docs.

src/bptl/work_units/valid_sign/tasks.py
```diff
@@ -0,0 +1,382 @@
...
+                    charset="utf8",
```
sergei-maertens, line 198: this is not something you can assume - there's even no guarantee at all that the document is text based. a simple `application/pdf` mime-type already does not make sense with a charset declaration.

src/bptl/work_units/valid_sign/tasks.py
```diff
@@ -0,0 +1,382 @@
...
+                )
+                for line in response.iter_lines():
```
sergei-maertens, line 200: you should be dealing with binary data, not strings, since the document mime type can be anything (think docx, pdf, audio files...)

I've used the following snippet in ibabs jail project:

```py
class wrapper:
    def __init__(self, response, chunk_size=4096):
        self.response = response
        self.chunk_size = chunk_size

    def __iter__(self):
        for chunk in self.response.iter_content(chunk_size=self.chunk_size):
            yield chunk
```

src/bptl/work_units/valid_sign/tasks.py
```diff
@@ -0,0 +1,382 @@
...
+            document_size = int(response.headers["content-length"]) / 1e6
+
+            # If the size of the document is above the max size or if all the documents together have already reached
+            # the maximum size, write the file content to a temporary file
+            if (
+                document_size > settings.MAX_DOCUMENT_SIZE
+                or (current_total_documents_size + document_size)
+                > settings.MAX_TOT_DOCUMENT_SIZE
+            ):
+                tmp_file_object = TemporaryUploadedFile(
```
sergei-maertens, line 194: this file should be opened in binary mode before writing, not sure what the default is

src/bptl/work_units/valid_sign/tasks.py
```diff
@@ -0,0 +1,382 @@
...
+                    content_type="application/octet-stream",
+                    size=sys.getsizeof(response.content),
+                    charset="utf8",
+                )
+                for line in response.iter_lines():
+                    tmp_file_object.write(line)
+                tmp_file_object.flush()
+                doc_tuple = (document_data["titel"], tmp_file_object)
+            else:
+                doc_tuple = (document_data["titel"], response.content)
```
sergei-maertens, line 205: I'd wrap `response.content` in a `BytesIO` object, so that further down everything is a file like object with `read()` method etc.

src/bptl/work_units/valid_sign/tasks.py
```diff
@@ -0,0 +1,382 @@
...
+
+        signers = self._get_signers_from_package(package)
+        approvals = self._get_approvals(signers)
+
+        attached_documents = []
+        for doc_name, doc_content in documents:
+            url = f"{validsign_client.base_url}api/packages/{package['id']}/documents"
+            payload = {"name": doc_name, "extract": True, "approvals": approvals}
+            body = {"payload": json.dumps(payload)}
+            if isinstance(doc_content, TemporaryUploadedFile):
```
sergei-maertens, line 302: if you wrap the document data in `BytesIO` above, you don't need this `isinstance` check

sergei-maertens (changes requested):

src/bptl/work_units/valid_sign/tasks.py
SilviaAmAm: :facepalm:

sergei-maertens (approved):

PR #72, opened 2020-08-24
Feature/task create rol

annashamray:


src/bptl/work_units/zgw/tasks/rol.py
```diff
@@ -0,0 +1,81 @@
...
+    """
+    Create a new ROL for the ZAAK in the process.
+
+    **Required process variables**
+
+    * ``zaakUrl``: full URL of the ZAAK to create a new rol for
+    * ``omschrijving``: roltype.omschrijving for the ROL
+    * ``betrokkeneType``: type of the betrokkene
+    * ``roltoelichting``: description of the ROL
+    * ``betrokkene``: JSON object containing the data for ROL.betrokkeneIdentificatie. See
```
sergei-maertens, line 24: This is not what was specified in the ticket:

> `betrokkene`: a JSON object containing the rest of the body for rol_create

In this form, it's not possible to pass a URL for a `betrokkene`, as you are now enforcing everyone to use `betrokkeneIdentificatie` which should only be an escape hatch and not the de-facto way.

This is also not what happens in `bptl.work_units.zgw.tasks.zaak.CreateZaakTask.create_rol` as mentioned in the ticket.

`betrokkene` is _most_ of the body of `rol_create` operation, and by doing that, you don't need the variables `betrokkeneType` and `roltoelichting` either, since they are part of that body.

src/bptl/work_units/zgw/tasks/rol.py
```diff
@@ -0,0 +1,81 @@
...
+        variables = self.task.get_variables()
+
+        zrc_client = self.get_client(APITypes.zrc)
+        zaak_url = check_variable(variables, "zaakUrl")
+        zaak = zrc_client.retrieve("zaak", url=zaak_url)
+
+        ztc_client = self.get_client(APITypes.ztc)
+        query_params = {
+            "zaaktype": zaak["zaaktype"],
+            "omschrijvingGeneriek": check_variable(variables, "omschrijving"),
```
sergei-maertens, line 59: this is not what was specified in the ticket:

> `omschrijving`: the roltype omschrijving, use this in combination with zaak.zaaktype to find the correct rolType reference

`omschrijvingGeneriek` is not unique with regards to the zaaktype, but `omschrijving` is.

sergei-maertens (changes requested):

src/bptl/work_units/zgw/tasks/rol.py
annashamray: according to catalogi schema (https://openzaak.utrechtproeftuin.nl/catalogi/api/v1/schema/#operation/roltype_list) `omschrijving` is not in the list of available query params, but `omschrijvingGeneriek` is in the list. Should I filter on it?

src/bptl/work_units/zgw/tasks/rol.py
annashamray: The implementation is changed - now all required rol variables (except `zaak` and `roltype`) are retrieved from `betrokkene` input variable

src/bptl/work_units/zgw/tasks/rol.py
sergei-maertens: no, it can't be queried. best is to just do the filtering in Python

annashamray: roltypen filtering on `omschrijving` is added

src/bptl/work_units/zgw/tasks/rol.py
```diff
@@ -0,0 +1,82 @@
...
+            rol_type
+            for rol_type in rol_typen["results"]
+            if rol_type["omschrijving"] == omschrijving
+        ]
+        if not rol_typen:
+            raise ValueError(
+                f"No matching roltype with zaaktype = {zaak['zaaktype']} and omschrijving = {omschrijving} is found"
+            )
+
+        zrc_client = self.get_client(APITypes.zrc)
```
sergei-maertens, line 67: No need to get this client again, it's already available from line 48

sergei-maertens (approved):

src/bptl/work_units/zgw/tasks/rol.py
annashamray: the line is removed. Thanks for catching

sergei-maertens (approved):

Issue #71, opened 2020-08-24
Implement a work unit to lock and unlock a DRC document


sergei-maertens:
Upstream issue: https://github.com/GemeenteUtrecht/ZGW/issues/513

---

There should be BPTL tasks/work units to "freeze" a document by locking it, and also a task to unlock it again after locking it.

Acceptance criteria:

- [x] There is a document-lock work unit
    - [x] Input is a documenten API URL
    - [x] Second input is the `services` variable with credentials
    - [x] The result is that the document is locked in the API
    - [x] The result variable is the lock ID
    - [x] The work unit is documented properly and similar to the other tasks
- [x] There is a document-unlock work unit
    - [x] Input is a documenten API URL and the lock ID
    - [x] Third input is the `services` variable with credentials
    - [x] The result is that the document is unlocked in the API
    - [x] There are no result variables
    - [x] The work unit is documented properly and similar to the other tasks

sergei-maertens:
Fixed via #74

PR #74, opened 2020-09-04
Issue #71 Tasks to lock/unlock a document

SilviaAmAm:


src/bptl/work_units/documents/tasks.py
```diff
@@ -0,0 +1,147 @@
...
+        drc_client = self.build_client(document_url)
+        document_data = drc_client.retrieve(
+            resource="enkelvoudiginformatieobject",
+            url=document_url,
+        )
+
+        # Lock document
+        response = drc_client.operation(
+            operation_id="enkelvoudiginformatieobject_lock",
+            uuid=document_data["uuid"],
```
sergei-maertens, line 99: it doesn't look like `uuid` is part of the response body. Probably the best option now is to skip the `retrieve` call completely, and instead split the URL on `/` and grab the last part, use that as UUID. Make sure to use `urlparse` so you don't accidentally have querystring parameters like `versie=123` in there.

src/bptl/work_units/documents/tasks.py
```diff
@@ -0,0 +1,147 @@
...
+          }
+
+    **Sets no process variables**
+    """
+
+    def perform(self) -> dict:
+        variables = self.task.get_variables()
+
+        # Retrieve document
+        document_url = check_variable(variables, "document")
```
sergei-maertens, line 132:
```suggestion
        document_url = check_variable(variables, "informatieobject")
```

src/bptl/work_units/documents/tasks.py
```diff
@@ -0,0 +1,147 @@
...
+@register
+class UnlockDocumentTask(DocumentTask):
+    """Unlock a DRC document.
+
+    **Required process variables**
+
+    * ``document``: str, API URL of a DRC document.
+        The API must comply with the Documenten API 1.0.x (https://vng-realisatie.github.io/gemma-zaken/standaard/documenten/index).
+
+    * ``lockId``: str, Lock ID for the locked DRC document.
```
sergei-maertens, line 115:
```suggestion
    * ``lockId``: string, Lock ID for the locked DRC document.
```
Better to spell out as much as possible, this documentation is aimed at non-programmers and especially not python programmers.

sergei-maertens (changes requested):
These should be part of the `bptl.work_units.zgw` package rather than their own package.

sergei-maertens (approved):

Issue #76, opened 2020-09-10
Introduce multiple queues


sergei-maertens:
Celery's basic config is having a single queue named `celery` where all jobs are submitted.

Since the introduction of long polling to Camunda in 6d03a3dc7506b2a06cafef762859254f7585921b this introduces a risk - if the prefetch multiplication is mis-configured, tasks that should run instantly can be queued inside of a worker behind a long-poll task (which may run for up to 10 minutes in its current config).

To prevent mistakes like that (and being able to ACK early) - it makes sense to have two queues:

* a queue for long-poll jobs and only those sort of jobs
* a queue for work that should be processed ASAP

This also means that workers must be spun up/assigned to a particular queue (by default they listen to all queues).

The desired configuration YAML in infrastructure should look like:

```yaml
django_app_k8s_celery_workers:
    - queue: celery
      replicas: 3
    - queue: long-polling
      replicas: 2
```

Acceptance criteria:

- [x] The settings define a second queue dedicated to long-polling
- [x] The long-polling related celery tasks publish to that dedicated queue
- [x] Other tasks publish to the default `celery` queue
- [x] Workers can be spun up for a particular queue
- [x] Infrastructure update
- [x] Update documentation on queues

PR #78, opened 2020-09-16
Introduced the long-polling queue

damm89:
The long-polling tasks will now enter a long-polling queue instead of the default queue. All other tasks enter the default queue.

TO DO: Infrastructure update
#76


src/bptl/conf/base.py
```diff
@@ -397,8 +399,30 @@
...
 CELERY_RESULT_BACKEND = os.getenv("CELERY_RESULT_BACKEND", "redis://localhost:6379/0")
+
 # Add a 30 minutes timeout to all Celery tasks.
 CELERY_TASK_SOFT_TIME_LIMIT = 30 * 60
+
+# Setup Celery queues: default & long
+CELERY_DEFAULT_QUEUE = "default"
+CELERY_DEFAULT_EXCHANGE = "default"
+CELERY_DEFAULT_EXCHANGE_TYPE = "direct"
+CELERY_DEFAULT_ROUTING_KEY = "default"
```
sergei-maertens, line 410: I'm not sure we want to override this - we use redis for message broker, result store and as cache backend. They *should* each be using a default database inside of the redis server, but on the off-chance that doesn't happen I'd like to avoid seeing keys with "default" and rather stick to the default value of "celery".

src/bptl/camunda/tasks.py
```diff
@@ -29,6 +29,8 @@
             (settings.LONG_POLLING_TIMEOUT_MINUTES * 60) + 1
         ),  # timeout if something goes wrong, in seconds
     },
+    exchange="long-polling",
+    queue="long-polling",
```
sergei-maertens, line 33: checking the docs at https://docs.celeryproject.org/en/4.4.0/userguide/tasks.html#general it seems like the `queue` option for the `@app.task` decorator is no longer available, and the routing setting should instead be used.

```suggestion
```

I'm a little bit confused by https://docs.celeryproject.org/en/4.4.0/userguide/routing.html#specifying-task-destination though, where 2. seems to imply that the `queue` task option _does_ still work.

src/bptl/conf/base.py
```diff
@@ -397,8 +399,30 @@
...
+)
+
+# Setup Celery routes for long-polling
+CELERY_TASK_ROUTES = {
+    "bptl.camunda.tasks.task_fetch_and_lock": {
+        "queue": "long-polling",
+        "exchange": "long-polling",
+        "routing_key": "long-polling",
+    }
+}
```
sergei-maertens, line 424: The docs seem to advise against specifying this explicitly and instead relying on the automatic routing: https://docs.celeryproject.org/en/4.4.0/userguide/routing.html#routing-automatic

So this would then become something like:

```suggestion
# Setup Celery routes for long-polling
CELERY_TASK_ROUTES = {
    "bptl.camunda.tasks.task_fetch_and_lock": {
        "queue": "long-polling",
    },
    "bptl.camunda.tasks.task_schedule_new_fetch_and_lock": {
        "queue": "long-polling",
    },
}
```

I think this nicely hides away the Kombu/Exchange configuration where the default is fine for us.

src/bptl/conf/base.py
```diff
@@ -407,8 +431,14 @@
         # start the polling after 10 seconds, and keep it running. The task itself
         # is marked graceful, so it'll just return None and not be scheduled again.
         "schedule": schedule(run_every=10),
+        "options": {
+            "queue": "long-polling",
+            "exchange": "long-polling",
+            "routing_key": "long-polling",
+        },
```
sergei-maertens, line 438: I'm not sure if we need this if the `CELERY_TASK_ROUTES` are defined :thinking:

sergei-maertens (changes requested):
General feeling is that a lot of configuration is now duplicated with the same effect, but not sure which one overrules what. I *think* we can reduce that down mostly into the `CELERY_TASK_ROUTES` setting, but will of course require some testing.

Additionally, `src/celerybeat-schedule` should not be in version control, I think it's a file generated from the celery beat setting and should be `.gitignore`'d.

bin/celery_worker.sh
```diff
@@ -2,13 +2,28 @@
...
 
+while [ $# -gt 0 ]; do
+
+   if [[ $1 == *"--"* ]]; then
+        param="${1/--/}"
+        declare $param="$2"
+   fi
+
+  shift
+done
```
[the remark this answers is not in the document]
damm89: Was voor dev purpose om ff snel variabelen door te kunnen geven

src/bptl/conf/base.py
```diff
@@ -190,7 +191,8 @@
 STATICFILES_DIRS = (os.path.join(DJANGO_PROJECT_DIR, "static"),)
 
 # List of finder classes that know how to find static files in
-# various locations.
+# various locations.    apply_async()
```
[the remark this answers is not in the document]
damm89: Woops

src/bptl/conf/base.py
damm89: Default van celery is prima, maar als je niet de automatic routing gebruikt (die ik nergens zag staan en dus maar aannam dat we daar niet op "vertrouwden") was het voor mij onduidelijk of de ongeroute tasks wel naar de default queue zouden gaan of dat ze ook in de long-polling queue terecht zouden kunnen komen

src/bptl/camunda/tasks.py
damm89: Me2 :X

src/bptl/conf/base.py
damm89: Prima - dan zetten we dus automatic routing aan.

src/bptl/conf/base.py
sergei-maertens: Yep, dat deden ze inderdaad. In mijn POC branch kan je dat ook nagaan m.b.v. flower

sergei-maertens (changes requested):
Couple more remarks left!

Please add the following files to `.gitignore`:

- `celerybeat-schedule`
- `celerybeat.pid`

I'm missing the possibility now to pass the queue name(s) to listen to for the `bin/celery_worker.sh` script. Also the `--name` parameter, derived from the queue name was convenient, but that's going to give issues when deployed with multiple replicas (and then they have all the same names). Though we can probably use the pod name as part of the name passed to the worker. In short - it should be possible to explicitly define the name, if not done so -> generate a name based on the requested queue(s).

sergei-maertens (approved):

PR #83, opened 2020-10-09
Feature/send email

damm89:
Added a generic email work unit.
GemeenteUtrecht/ZGW#558
GemeenteUtrecht/ZGW#538


sergei-maertens (commented):
Patterns look good, will review in more detail tomorrow!

src/bptl/work_units/email/serializers.py
```diff
@@ -0,0 +1,37 @@
+from django.utils.translation import ugettext_lazy as _
+
+from djchoices import ChoiceItem, DjangoChoices
+from rest_framework import serializers
+
+
+class ValidTemplateChoices(DjangoChoices):
+    generiek = ChoiceItem("generiek", "../templates/generic_email.txt")
+    accordering = ChoiceItem("accordering", "../templates/review.txt")
+    advies = ChoiceItem("advies", "../templates/review.txt")
```
sergei-maertens, line 10: relative paths look weird, especially because django has template resolving already based on app name.

`django.template.loader.get_template` will search according to the config in the `TEMPLATES` setting.

Especially within a docker context where you can mount volumes containing custom templates, this is relevant for broader adoption than just the templates in this codebase.

I'd change this into `email/mails/generic.txt` and `email/mails/review.txt` rather.

I'm also not sure that using `DjangoChoices` here makes sense - because you're using the same label for different values which makes it impossible in user interfaces to select the correct choice.

Shouldn't this rather be a simple mapping?

```suggestion
MAIL_TEMPLATES = {
    "generiek": "email/mails/generic.txt",
    "accordering": "email/mails/review.txt",
    "advies": "email/mails/review.txt",
}
```

and then the serializer fields becomes something rather like:

```py
template = serializers.ChoiceField(choices=((key, key) for key in MAIL_TEMPLATES)
```

src/bptl/work_units/email/serializers.py
```diff
@@ -0,0 +1,37 @@
...
+
+class SendEmailSerializer(serializers.Serializer):
+    sender = EmailPersonSerializer()
+    receiver = EmailPersonSerializer()
+    email = EmailSerializer()
+    template = serializers.ChoiceField(
+        choices=[key for key in ValidTemplateChoices.labels],
+        required=True,
+    )
+    kwargs = KwargsSerializer()
```
sergei-maertens, line 37: I'd name this `context` rather than `kwargs`

src/bptl/work_units/email/tasks.py
```diff
@@ -0,0 +1,89 @@
...
+    * ``template``: string with template name. Choices:
+        .. code-block:: list
+                [
+                    generiek, accordering, advies,
+                ]
+
+    **Optional process variables**
+
+    * ``kownslFrontendUrl``: string with full URL for end users to submit their review.
+                Valid choices are currently: generic/approval/advice.
```
sergei-maertens, line 52: I don't see how this is a valid choice for a URL? :P

src/bptl/work_units/email/tasks.py
```diff
@@ -0,0 +1,89 @@
...
+    * ``reminder``: boolean.
+    * ``deadline``: datetime: "2020-04-20 16:20:00".
```
sergei-maertens, line 54: Can we make the examples timezone-aware?

```suggestion
    * ``deadline``: datetime: "2020-04-20T16:20:00+00:00".
```

src/bptl/work_units/email/tasks.py
```diff
@@ -0,0 +1,89 @@
...
+    """
+
+    def perform(self):
+        variables = self.task.get_variables()
+        send_email = SendEmailSerializer(data=variables)
+        send_email.is_valid(raise_exception=True)
```
sergei-maertens, line 60: I like the usage of using a serializer for validation, but now the errors will look different from what people are using in the Camunda incident logging and the BPTL exception details.

Can you somehow turn this into `bptl.tasks.base.MissingVariable` exceptions where relevant? This is why `bptl.tasks.base.check_variable` exists.

```suggestion
        send_email = SendEmailSerializer(data=variables)
        send_email.is_valid(raise_exception=True)
```

src/bptl/work_units/email/tasks.py
```diff
@@ -0,0 +1,89 @@
...
+            send_email["email"][
+                "subject"
+            ] = f"HERINNERING: {send_email['email']['subject']}"
+
+        # Render and send
+        email_message = email_template.render(email_context)
+        send_mail(
+            subject=send_email["email"]["subject"],
+            message=email_message,
+            from_email=settings.DEFAULT_FROM_EMAIL,
```
sergei-maertens, line 87: This should either use the sender e-mail or be configurable in another way, I think.

Problem with using sender-email is that you get SPF and DKIM etc. kicking in, so for that you'd at least have to use the `reply_to` set to the sender address. I'd go for that addition for MVP and stick to DEFAULT_FROM_EMAIL for now.

src/bptl/work_units/email/templates/review.txt
```diff
@@ -0,0 +1,6 @@
+{% extends "base.txt" %}
+{% load i18n %}{% block body %}{% if reminder %}HERINNERING: {% endif %}Uw {{ review_type }} is vereist. {% if deadline %}U heeft tot {{ deadline }} om te reageren.{% endif %}
+
+Klik alstublieft <a href="{{ kownslFrontendUrl }}">hier</a>.
```
sergei-maertens, line 4: I'm not sure how office 365 treats this (I think GMail does parse it as HTML), but having HTML in a .txt file/template seems odd. The e-mail sending code also sends it as plain text, and has no HTML alternative attached. Either way, I would make sure the full URL is just visible in the mail, and not hidden behind a "hier" text as an element attribute. If clients strip tags, the URL is completely gone and can't even be pasted by recipients.

For future reference, I expect that there will exist some sort of template/"sjabloon" thing that will spit out the actual mail content (probably as HTML) which will facilitate the actual templating. Probably we're already doing too much templating here, and I do want to avoid building an entire mail templating solution in BPTL - that should be another (micro)service that possibly gets called in the process before sending the mail.

sergei-maertens (changes requested):
The `src/bptl/work_units/email/templates/__init__.py` file should not exist, the `templates` dir is not a python package.

Additionally, within a Django app it's best practice to namespace the templates with the app name, so put them under `src/bptl/work_units/email/templates/email/`. I'd also create a sub-directory in there called `mails` containing the mail templates, vs. possible future HTML templates for views.

src/bptl/work_units/email/serializers.py
damm89: fair enough

src/bptl/work_units/email/tasks.py
damm89: Oops!

src/bptl/work_units/email/tasks.py
damm89: absolutely. had something with check_variable first but I like serializers better... will find a middle ground solution

sergei-maertens: cool! it did give me the idea that we should probably layer our own validation on top of DRF serializers, but I don't expect this to be done before 1.0 - could be a 1.1 refactor thing and I'd appreciate input!

damm89: Right now I did an error catch and I'm matching the error_code to raise a MissingVariable error where relevant. It's definitely not ideal. 

In this case we want our validation to run before the DRF serializers validation (and its fields' validation), so custom (field/serializer) validators aren't useful because they get called much later on.

I was thinking in the direction of writing a custom serializer object with some class methods to check for required fields and recursively looking for the fields in the initial_data. Then we could have the relevant serializers inherit the is_valid function from our custom serializer. Would be nice to hear your arguments against. :-)

src/bptl/work_units/email/templates/review.txt
damm89: Sure thing.

src/bptl/work_units/email/tasks.py
sergei-maertens: I'm having a hard time picturing what you mean, can you make a gist with some small (pseudo) code examples to illustrate it? We can discuss on Slack then. There's also no urgency for this, so don't feel pressured to prio this!

sergei-maertens (approved):
Note to self: when deploying this, the outgoing mail-server needs to be hooked up in the infrastructure repo on Gitlab too.

PR #87, opened 2020-10-21
Added work unit to look up a zaak reference

sergei-maertens:
Work unit in the context of being able to relate an earlier case to a new case, e.g. for "bezwaar".

The envisioned flow is a form with the following inputs:

- text field to enter Zaak reference
- any other fields required to create the bezwaar zaak

The process model then _could_ contain the following tasks:

1. Create zaak (bezwaarzaak)
2. Look up the Zaak reference
3. if the reference resolves, relate the looked up zaak and the bezwaarzaak

### DIFF
--- a/src/bptl/work_units/zgw/tasks/zaak.py
+++ b/src/bptl/work_units/zgw/tasks/zaak.py
@@ -6,6 +6,7 @@ from django.utils import timezone
 
 from zgw_consumers.constants import APITypes
 
+from bptl.tasks.base import check_variable
 from bptl.tasks.registry import register
 
 from ..nlx import get_nlx_headers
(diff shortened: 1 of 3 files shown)


src/bptl/work_units/zgw/tasks/zaak.py
```diff
@@ -231,3 +232,61 @@ def perform(self):
...
+
+        .. code-block:: json
+
+          {
+              "<zrc alias>": {"jwt": "Bearer <JWT value>"},
+          }
+
+    **Optional process variables (Camunda exclusive)**
+
+    * ``callbackUrl``: send an empty POST request to this URL to signal completion
```
damm89, line 263: Why is the callbackUrl needed/handy/used?

src/bptl/work_units/zgw/tasks/zaak.py
```diff
@@ -231,3 +232,61 @@ def perform(self):
...
+                "bronorganisatie": bronorganisatie,
+                "identificatie": identificatie,
+            },
+        )
+
+        # paginated response
+        if zaken["results"]:
+            zaak_url = zaken["results"][0]["url"]
+        else:
+            zaak_url = None
```
damm89, line 290: Should None not be an empty string since zaak_url is a string normally?

damm89 (commented):
Rest looks good.

src/bptl/work_units/zgw/tasks/zaak.py
sergei-maertens: This was a request from another party looking into using BPTL. They found it not ideal that they'd have to poll the Camunda process instance to check if the action was completed, so they can provide an optional callback URL that will be notified on completion of the Camunda task.

I don't think it's actually used, but it's a generic implementation available for any task running in a Camunda context (i.e. -> not used if you call the task via the REST API).

I'm considering dropping that support in 1.0 because it's not used and because we're using long-polling now that should reduce latency significantly.

src/bptl/work_units/zgw/tasks/zaak.py
sergei-maertens: Discussed on Slack, we both can't give a hard reason why it should be one or the other.

Checked with Java's interpretation/meaning of `null` and it matches the intended meaning, i.e. the object does not exist. However, changing types in Python is kinda iffy. We'll see if we run into problems or not - maybe down the road with the Camunda Expression Language there will be a clear reason to pick one or the other.

damm89 (approved):

PR #88, opened 2020-10-26
Feature/send email

damm89:
Added tasks in kownsl work_unit that allows for retrieval of the response status of a review request such as approval and advice. Also implemented the feature that allows for retrieval of deadlines per set of kownslUsers.


src/bptl/work_units/kownsl/tasks.py
```diff
@@ -123,3 +146,112 @@ def get_approval_status(task: BaseTask) -> dict:
...
+    """
+    Get the reviewers who have not yet responded to a review request so that
+    a reminder email can be sent to them if they exist.
+
+    In the task binding, the service with alias ``kownsl`` must be connected, so that
+    this task knows which endpoints to contact.
+
+    **Required process variables**
+
+    * ``reviewRequestId``: the identifier of the Kowns review request, used to update
```
sergei-maertens, line 162:
```suggestion
    * ``reviewRequestId``: the identifier of the Kownsl review request, used to update
```

sergei-maertens: this documentation in general seems to be a bit off - there's no `update` happening, right?

src/bptl/work_units/kownsl/tasks.py
```diff
@@ -123,3 +146,112 @@ def get_approval_status(task: BaseTask) -> dict:
...
+      the object in the API.
+    * ``hoofdZaakUrl``: URL reference to the zaak used for the review itself.
```
sergei-maertens, line 164: I think this was an old, deprecated approach where a zaak was created for the review itself. Should probably be `zaakUrl` instead.

src/bptl/work_units/kownsl/tasks.py
```diff
@@ -123,3 +146,112 @@ def get_approval_status(task: BaseTask) -> dict:
...
+
+    **Sets the process variables**
+
+    * ``remindThese``: a JSON-object containing a list of usernames who need reminding:
+
+      .. code-block:: json
+            [
+                "user1",
+                "user2",
+            ]
```
sergei-maertens, line 175:
```suggestion
      .. code-block:: json

            [
                "user1",
                "user2",
            ]
```
code-formatting in RST is iffy

src/bptl/work_units/kownsl/tasks.py
```diff
@@ -123,3 +146,112 @@ def get_approval_status(task: BaseTask) -> dict:
...
+    """
+
+    # Get the review request with id as given in variables
+    review_request = get_review_request(task)
+
+    review_request_id = review_request.get("id")
```
sergei-maertens, line 181:
```suggestion
    review_request_id = review_request["id"]
```
you expect it to be from the Kownsl API, so the `id` attribute is set, always. If it isn't, that's a hard error from the API that should raise a `KeyError`

src/bptl/work_units/kownsl/tasks.py
```diff
@@ -123,3 +146,112 @@ def get_approval_status(task: BaseTask) -> dict:
...
+
+    # Get review request type to set operation_id
+    review_type = review_request.get("review_type")
```
sergei-maertens, line 184:
```suggestion
    review_type = review_request["review_type"]
```

src/bptl/work_units/kownsl/tasks.py
```diff
@@ -123,3 +146,112 @@ def get_approval_status(task: BaseTask) -> dict:
...
+    The returned value is the deadline minus one day.
+
+    In the task binding, the service with alias ``kownsl`` must be connected, so that
+    this task knows which endpoints to contact.
+
+    **Required process variables**
+
+    * ``reviewRequestId``: the identifier of the Kowns review request, used to update
+      the object in the API.
+    * ``hoofdZaakUrl``: URL reference to the zaak used for the review itself.
```
sergei-maertens, line 233: See earlier remarks

src/bptl/work_units/kownsl/tasks.py
```diff
@@ -123,3 +146,112 @@ def get_approval_status(task: BaseTask) -> dict:
...
+    """
+    Get the reminder for the set of reviewers who are requested.
+    The returned value is the deadline minus one day.
+
+    In the task binding, the service with alias ``kownsl`` must be connected, so that
+    this task knows which endpoints to contact.
+
+    **Required process variables**
+
+    * ``reviewRequestId``: the identifier of the Kowns review request, used to update
```
sergei-maertens, line 231: See earlier remarks w/r to the "update" part of the docstring

src/bptl/work_units/kownsl/tests/tests_kownsl_tasks.py
```diff
@@ -0,0 +1,183 @@
...
+        cls.task_dict = {
+            "topic_name": "some-topic",
+            "worker_id": "test-worker-id",
+            "task_id": "test-task-id",
+            "variables": {
+                "hoofdZaakUrl": {
+                    "type": "String",
+                    "value": "https://zaken.nl/api/v1/zaak/123",
+                    "valueInfo": {},
+                },
```
sergei-maertens, line 60: you can/should use [`django_camunda.utils.serialize_variable`](https://github.com/maykinmedia/django-camunda/blob/master/django_camunda/utils.py#L69) for this instead

src/bptl/work_units/zac/models.py
```diff
@@ -0,0 +1,42 @@
...
+        help_text=_("HTTP Authorization header name, required if the API is not open."),
+    )
+    header_value = models.CharField(
+        _("header value"),
+        max_length=255,
+        blank=True,
+        help_text=_(
+            "HTTP Authorization header value, required if the API is not open."
+        ),
+    )
```
sergei-maertens, line 29: Hrm, I'd rather see a FK to `zgw_consumers.Service` for this instead, no? You already configure the API root in that model + the auth headers. OTOH that's also how BRP is done now, so probably we should just clean it up all in one go before 1.0. I've created #89 for that.

src/bptl/work_units/zac/tasks.py
```diff
@@ -0,0 +1,70 @@
...
+from .serializers import ZacUsersDetailsSerializer
+
+
+@register
+class UserDetailsTask(WorkUnit):
+    """
+    Requests email and name data from usernames from the zac.
+
+    In the camunda process models accorderen/adviseren we have a list of usernames from
+    the zac. In order to send signalling emails, we will need to fetch
```
sergei-maertens, line 18:
```suggestion
    the zac. In order to send signaling emails, we will need to fetch
```

src/bptl/work_units/zac/tasks.py
```diff
@@ -0,0 +1,70 @@
...
+    the email addresses and names from the zac and feed it back to the camunda process.
+
+    In this first implementation a simple and direct get request is done
+    at the zac.accounts.api endpoint.
+
+    The task requires the service with alias "zac" to be configured in the work_unit
+    services and be registered in the ZAC where the token will be set.
```
sergei-maertens, line 25: Currently not the case because no `zgw_consumers.Service` is used, rather it's using the static configuration.

src/bptl/work_units/zac/tasks.py
```diff
@@ -0,0 +1,70 @@
...
+
+    **Required process variables**
+
+    * ``usernames``: JSON with usernames.
+        .. code-block:: json
+                [
+                    "user1",
+                    "user2",
+                    "user3",
+                ]
```
sergei-maertens, line 35:
```suggestion
    * ``usernames``: JSON with usernames.
        .. code-block:: json
        
                [
                    "user1",
                    "user2",
                    "user3",
                ]
```

src/bptl/work_units/zac/tasks.py
```diff
@@ -0,0 +1,70 @@
...
+
+    * ``userData``: a JSON-object containing a list of user names and emails:
+
+      .. code-block:: json
+            [
+                {
+                    "name": "FirstName LastName"
+                    "email": "test@test.nl"
+                },
+            ]
```
sergei-maertens, line 47:
```suggestion
      .. code-block:: json
      
            [
                {
                    "name": "FirstName LastName"
                    "email": "test@test.nl"
                }
            ]
```

src/bptl/work_units/zac/tasks.py
```diff
@@ -0,0 +1,70 @@
...
+
+    """
+
+    def get_client_response(self) -> Response:
+        client = ZACClient()
+        variables = self.task.get_variables()
+        usernames = check_variable(variables, "usernames")
+        url = f"accounts/api/users?filter_users={','.join(usernames)}&include=True"
+        response = client.get(url)
```
sergei-maertens, line 56: I really dislike building querystrings by hand:

- no escaping of inputs happened
- not readable

Please rely on `requests`' abstraction - pass in the query parameters as a dictionary to the underlying `requests.get` call, and let requests handle the querystring/URL building with the appropriate URL quoting.

Also, what happens if some wacky username contains a comma? :thinking: 

The typical approach for multi-dimensional query string arguments is to repeat the argument:

`?username=foo&username=bar&username=baz` - I'm pretty sure [django-filter](https://django-filter.readthedocs.io/en/stable/) can handle that.

src/bptl/work_units/zac/tasks.py
```diff
@@ -0,0 +1,70 @@
...
+        serializer.is_valid(raise_exception=True)
+        return serializer.data
+
+    def perform(self) -> dict:
+        response = self.get_client_response()
+        validated_data = self.validate_data(response.json())
+        validated_data = json.dumps(validated_data["results"])
+        return {
+            "userData": json.loads(validated_data),
+        }
```
sergei-maertens, line 70: This looks odd to me - what's the point of json dumping and then immediately loading it again?

```suggestion
        return {
            "userData": validated_data["results"],
        }
```

sergei-maertens (changes requested):

src/bptl/work_units/kownsl/tasks.py
```diff
@@ -20,6 +22,27 @@ def get_client(task: BaseTask, alias: str = "kownsl") -> ZGWClient:
     return client
 
 
+def get_review_request(task: BaseTask) -> dict:
+    """
+    Get a single review request from kownsl.
+    """
+    variables = task.get_variables()
+
+    zaak_url = check_variable(variables, "hoofdZaakUrl")
```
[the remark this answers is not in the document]
damm89: Since this is in the accordering sub process the zaakUrl is not available.

src/bptl/work_units/kownsl/tasks.py
damm89: Youre right. Dumb copy paste error.

src/bptl/work_units/kownsl/tasks.py
damm89: It's still called hoofdzaakurl in accordering sub process.

src/bptl/work_units/kownsl/tests/tests_kownsl_tasks.py
damm89: for which part exactly?

src/bptl/work_units/zac/models.py
damm89: Okay. Leaving it in for now then.

src/bptl/work_units/zac/tasks.py
damm89: British/American spellings are always tricky :-) (Mother is British, wife is American)

src/bptl/work_units/zac/tasks.py
damm89: I suggested that to Kelvin as well but since we're not doing that from the frontend either (check Kelvin's implementation) I'd rather stay consistent...

src/bptl/work_units/zac/tasks.py
damm89: I don't like ordered dicts, but I wanted the dictionary to return a dict and not a string to stay consistent with expectations.

src/bptl/work_units/kownsl/tasks.py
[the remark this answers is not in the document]
damm89: Mmm, I see that approval status is also using zaakUrl. We're using the variables for the sub process to map zaakUrl to hoofdZaakUrl though in both the accordering and advisering sub processes. 

Align everything...?

src/bptl/work_units/zac/tasks.py
damm89: Done

src/bptl/work_units/kownsl/tasks.py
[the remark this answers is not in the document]
sergei-maertens: the BPTL work units should not assume anything about sub-process, for all they are aware, they run in the main process.

I prefer consistency in the work units, so `zaakUrl` everywhere. If this is being called in the context of a sub-process, it's up to the process modeller to map `hoofdZaakUrl` to `zaakUrl` for this particular service task. In Camunda, this can be done using the input/output parameters.

src/bptl/work_units/kownsl/tasks.py
sergei-maertens: same remark as https://github.com/GemeenteUtrecht/bptl/pull/88#discussion_r513497243

src/bptl/work_units/zac/tasks.py
sergei-maertens: > I suggested that to Kelvin as well but since we're not doing that from the frontend either (check Kelvin's implementation) I'd rather stay consistent...

we shall bring the frontend in line with this approach then :-) good "check" to keep in mind when reviewing frontend-code.

src/bptl/work_units/zac/tasks.py
sergei-maertens: hrm, we don't rely on ordering in dicts anywhere, as JSON has no inherent ordering for objects. The individual return variables are also set in arbitrary order by the process engine. `OrderedDict` is a `dict` subclass, and the public API called is `variables.items()`, so it shouldn't matter that you get `OrderedDict`, they quak like dict-ducks enough to serialize towards Camunda :)

If you really want to get rid of the `OrderedDict`, you should just use the `dict` constructor instead of abusing the json module:

```py
>>> from collections import OrderedDict
>>> example = OrderedDict((("foo", "bar"), ("fob", "baz")))
>>> example
OrderedDict([('foo', 'bar'), ('fob', 'baz')])
>>> dict(example)
{'foo': 'bar', 'fob': 'baz'}
```

src/bptl/work_units/kownsl/tasks.py
```diff
@@ -123,3 +144,194 @@ def get_approval_status(task: BaseTask) -> dict:
...
+    variables = task.get_variables()
+    kownsl_users = check_variable(variables, "kownslUsers")
+
+    # Get the review request with id as given in variables
+    review_request = get_review_request(task)
+    user_deadlines = review_request["user_deadlines"]
+
+    # Get deadline belonging to that specific set of kownslUsers
+    deadline_str = user_deadlines.get(kownsl_users[0])
+    deadline = datetime.datetime.strptime(deadline_str, "%Y-%m-%d").date()
```
sergei-maertens, line 248:
```suggestion
    deadline_str = user_deadlines[kownsl_users[0]]
    deadline = datetime.datetime.strptime(deadline_str, "%Y-%m-%d").date()
```

Because:

```py
>>> import datetime
>>> deadline_str = None
>>> datetime.datetime.strptime(deadline_str, "%Y-%m-%d").date()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: strptime() argument 1 must be str, not None
```

src/bptl/work_units/zac/tasks.py
```diff
@@ -0,0 +1,69 @@
...
+
+    """
+
+    def get_client_response(self) -> Response:
+        client = ZACClient()
+        variables = self.task.get_variables()
+        usernames = check_variable(variables, "usernames")
+        params = {"include": usernames}
+        response = client.get("accounts/api/users", params=params)
+        return response
```
sergei-maertens, line 56: I'd add a small sanity check in case the API has errors:

```suggestion
        response = client.get("accounts/api/users", params=params)
        response.raise_for_status()
        return response
```

src/bptl/work_units/kownsl/tests/tests_kownsl_tasks.py
sergei-maertens: basically the entire build-up of the task variables, as it abstracts the boring Camunda internals:

```py
cls.task_dict = {
    "topic_name": "some-topic",
    "worker_id": "test-worker-id",
    "task_id": "test-task-id",
    "variables": {
        "hoofdZaakUrl": serialize_variable("https://zaken.nl/api/v1/zaak/123"),
        "reviewRequestId": serialize_variable("1"),
    }
}
```

src/bptl/work_units/zac/tasks.py
damm89: Oh dear

sergei-maertens (approved):

PR #93, opened 2020-11-09
Ensure child/worker processes receive SIGTERM

sergei-maertens:
This wasn't the case before because child processes were spawned and
bash does not propagate the signals.

Especially with large grace times, having warm shutdown on workers work
correctly should keep zero-downtime deployments in kubernetes as fast
as possible.

### DIFF
--- a/bin/celery_beat.sh
+++ b/bin/celery_beat.sh
@@ -7,7 +7,7 @@ LOGLEVEL=${CELERY_LOGLEVEL:-INFO}
 mkdir -p celerybeat
 
 echo "Starting celery beat"
-celery beat \
+exec celery beat \
     --app bptl \
     -l $LOGLEVEL \
     --workdir src \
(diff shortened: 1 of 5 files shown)


damm89 (commented):
In our current setup - is python the first application that's run within the container? I'm guessing(?) it is. We could also run a dumb-init process before to handle signals?

sergei-maertens:
> In our current setup - is python the first application that's run within the container? I'm guessing(?) it is. We could also run a dumb-init process before to handle signals?

These are the `CMD` entries in the `Dockerfile` and/or k8s manifests. These shell scripts run as PID 1 - so when they reach the `exec` line, then `celery` / `uwsgi` / `flower` become PID 1 instead of bash.

Running an init process is not really the Docker way? In Docker you're supposed to run only a single process, in the foreground, on PID 1 as the container runtime sends signals to PID 1

damm89 (approved):

PR #94, opened 2020-11-11
Add support for setting Kownsl metadata

sergei-maertens:
This is required in our processes for when:

* user arrives at Kownsl page from e-mail
* Kownsl sends notification of review
* ZAC listens to notification
* ZAC fetches meta-information from Kownsl to be able to close the Camunda user task

This functionality allows us to update the meta-information that ZAC needs, from the process with an automated service task.

### DIFF
--- a/src/bptl/work_units/kownsl/tasks.py
+++ b/src/bptl/work_units/kownsl/tasks.py
@@ -230,6 +230,7 @@ def get_email_details(task: BaseTask) -> dict:
     **Sets the process variables**
 
     * ``email``: a JSON that holds the email content and subject.
+
       .. code-block:: json
 
             {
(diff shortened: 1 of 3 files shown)


src/bptl/work_units/kownsl/tasks.py
```diff
@@ -335,3 +295,40 @@ def get_email_details(task: BaseTask) -> dict:
...
+    metadata = check_variable(variables, "metadata")
+
+    client = get_client(task)
+    client.partial_update(
+        "reviewrequest",
+        data={"metadata": metadata},
+        uuid=review_request_id,
+    )
+
+    return {}
```
damm89, line 334: Why does this return an empty dict?

damm89 (approved):

annashamray (approved):

src/bptl/work_units/kownsl/tasks.py
sergei-maertens: Because a task must always return a dict, that's the contract of the task. The return value of the task is set as process variables, but this task does not set any process variables at all.

PR #96, opened 2020-11-16
Fixes #58 -- rework the auth/credentials system

### LINKED ISSUE #58: Rework the on-behalf-of credentials system

sergei-maertens:
- [x] Implement credentials store
- [x] Implement declaring required services (with aliases)
- [x] Refactor work units to use credentials store
    - [x] Keep `services` process var backwards compatible for now (deprecate, remove in 1.1)
    - [x]  BRP
    - [x] Camunda API
    - [x] Email
    - [x] Kadaster
    - [x] Kownsl
    - [x] ValidSign
    - [x] ZAC
    - [x] ZGW
- [x] Integrate with Autorisaties API for `App.app_id`
- [x] Refactor/cleanup some models/client building code


src/bptl/credentials/forms.py
```diff
@@ -0,0 +1,95 @@
...
+            applications = executor.map(_get_applications, autorisatie_apis)
+
+        all_apps = sum(applications, [])
+        all_apps = sorted(all_apps, key=lambda app: app["label"])
+
+        # if there's no data, no point in showing the field at all
+        if not all_apps:
+            del self.fields["autorisaties_application"]
+
+        self.fields["autorisaties_application"].choices = BLANK_CHOICE_DASH + [
```
damm89, line 57: lines 57-64 should be indented

src/bptl/credentials/forms.py
```diff
@@ -0,0 +1,95 @@
...
+
+    def clean(self):
+        super().clean()
+
+        app = self.cleaned_data.get("autorisaties_application")
+        label = self.cleaned_data.get("label")
+        app_id = self.cleaned_data.get("app_id")
+
+        # if there's no application specified that we can derive the values from, the
+        # other fields become required fields
```
damm89, line 74:
```suggestion
        # the other fields become required fields if there's no application specified 
        # that we can derive the values from
```
Slightly confusing description.

src/bptl/credentials/forms.py
```diff
@@ -0,0 +1,95 @@
...
+
+        # if there's no data, no point in showing the field at all
+        if not all_apps:
+            del self.fields["autorisaties_application"]
+
+        self.fields["autorisaties_application"].choices = BLANK_CHOICE_DASH + [
+            (application["url"], application["label"]) for application in all_apps
+        ]
+        self.fields["label"].required = False
+        self.fields["app_id"].required = False
```
damm89, line 61: This should only be False if there is app data?

src/bptl/credentials/forms.py
```diff
@@ -0,0 +1,95 @@
...
+
+        if self.instance.pk:
```
damm89, line 63: if there is no field, can there be an initial of that field?

src/bptl/credentials/forms.py
damm89: Mmm, it doesn't look like it's as easy as just indenting.

src/bptl/credentials/models.py
```diff
@@ -0,0 +1,143 @@
...
+                    _(
+                        "The service has '{auth_type}' type of authorization. You must fill "
+                        "out the header key and value."
+                    ).format(auth_type=self.service.get_auth_type_display()),
+                    code="incomplete-auth",
+                )
+            else:
+                logger.warning("Unknown service auth_type specified: %s", auth_type)
+
+    def get_auth_headers(self) -> Dict[str, str]:
```
damm89, line 115:
```suggestion
    def get_auth_headers(self) -> Union[Dict, Dict[str, str]]:
```
O.i.d.? Auth_header kan ook lege dict zijn toch?

src/bptl/credentials/forms.py
sergei-maertens: right, good catch. Need to take a good look at this!

src/bptl/credentials/forms.py
sergei-maertens: that's handled by the `def clean` - validation errors are raised if there is no app data.

src/bptl/credentials/forms.py
sergei-maertens: the DB value is always loaded as initial for `label` and `app_id`. We "conveniently" use the auth API identifier as the app_id, but technically app_id can be any string.

There is no DB field tracking which auth API application this points to, so there can only be an initial value if we have an existing DB record.

src/bptl/credentials/models.py
sergei-maertens: nope, the construction `Dict[str, str]` is essentially: `Dict[type_key, type_value]`. An empty dict still matches that type-hint, there just are no keys.

src/bptl/tasks/forms.py
```diff
@@ -71,6 +70,99 @@ def __init__(self, *args, **kwargs):
...
+        # selected callback - this is not validated because we don't have access to the
+        # form.
+        # TODO: check if django-extra-views CreateWithInlines has a proper solution for
+        # this
+        callback = self.data["callback"]
+        if not callback:
+            return  # won't validate the main form anyway
+
+        try:
+            task = register[self.data["callback"]]
```
damm89, line 86:
```suggestion
            task = register[callback]
```

src/bptl/tasks/forms.py
```diff
@@ -71,6 +70,99 @@ def __init__(self, *args, **kwargs):
...
+                form = self.forms[index]
+                form.add_error(
+                    "service",
+                    _(
+                        "The service for alias '{alias}' must be a '{api_type}' service."
+                    ).format(
+                        alias=required_service.alias,
+                        api_type=APITypes.labels[required_service.service_type],
+                    ),
+                )
```
damm89, line 131:
```suggestion
            alias = required_service.alias
            # if a particular alias is required, validate it as such
            if not alias:
                continue
            # check that the alias is present
            service_data = next(
                (
                    data
                    for data in self.cleaned_data
                    if data.get("alias") == alias
                ),
                None,
            )

            if service_data is None:
                raise forms.ValidationError(
                    _("Missing service alias '{alias}'").format(
                        alias=alias
                    )
                )

            _pinned_alias_services.append(service_data)

            # check that the service is of the right type
            api_type = service_data["service"].api_type
            if api_type != required_service.service_type:
                index = self.cleaned_data.index(service_data)
                form = self.forms[index]
                form.add_error(
                    "service",
                    _(
                        "The service for alias '{alias}' must be a '{api_type}' service."
                    ).format(
                        alias=alias,
                        api_type=APITypes.labels[required_service.service_type],
                    ),
                )
```
Just a bit more DRY.

src/bptl/tasks/forms.py
```diff
@@ -71,6 +70,99 @@ def __init__(self, *args, **kwargs):
...
+
+        # if any form errors are introduced, the formset is no longer valid and has no
+        # cleaned_data anymore -> so bypass the remaning validation
```
damm89, line 134:
```suggestion
        # cleaned_data anymore -> so bypass the remaining validation
```

src/bptl/tasks/forms.py
```diff
@@ -71,6 +70,99 @@ def __init__(self, *args, **kwargs):
...
+                for candidate in candidates
+            )
+            if not has_candidate:
+                raise forms.ValidationError(
+                    _(
+                        "Missing a service of type '{api_type}' which is required for this task."
+                    ).format(
+                        api_type=APITypes.labels[required_service.service_type],
+                    )
+                )
```
damm89, line 159:
```suggestion
            service_type = required_service.service_type
            has_candidate = any(
                candidate["service"].api_type == service_type
                for candidate in candidates
            )
            if not has_candidate:
                raise forms.ValidationError(
                    _(
                        "Missing a service of type '{api_type}' which is required for this task."
                    ).format(
                        api_type=APITypes.labels[service_type],
                    )
                )
```

src/bptl/tasks/forms.py
```diff
@@ -71,6 +70,99 @@ def __init__(self, *args, **kwargs):
         self.fields["alias"].required = False
 
 
+class BaseDefaultServiceFormset(BaseInlineFormSet):
+    def clean(self):
```
damm89, line 74: I feel like the clean function has an awful lot of logic in it... Any way to refactor this (later)?

src/bptl/work_units/kadaster/migrations/0001_create_brt_service.py
```diff
@@ -0,0 +1,53 @@
...
+from django.db import migrations
+
+from zgw_consumers.constants import APITypes, AuthTypes
+
+from bptl.tasks.registry import register
+
+from ..tasks import retrieve_openbare_ruimten
+from ..utils import ALIAS
+
+API_ROOT = "https://brt.basisregistraties.overheid.nl/api/v2"
```
annashamray, line 12: `zgw_consumers.Service` always adds a trailing slash to `api_root`
```suggestion
API_ROOT = "https://brt.basisregistraties.overheid.nl/api/v2/"
```

src/bptl/work_units/zac/migrations/0005_move_zac_config_to_default_service.py
```diff
@@ -0,0 +1,51 @@
...
+    if not ZACConfig.objects.filter(service__isnull=False).exists():
+        return
+
+    service = ZACConfig.objects.get().service
+    task_mappings = TaskMapping.objects.filter(
+        callback__startswith="bptl.work_units.zac.",
+    ).exclude(default_services=service)
+
+    # migrate the credentials to a placeholder app
+    app = App.objects.get_or_create(app_id="zac_MIGRATION_FIXME")
```
annashamray, line 22:
```suggestion
    app, created = App.objects.get_or_create(app_id="zac_MIGRATION_FIXME")
```

src/bptl/work_units/zac/migrations/0005_move_zac_config_to_default_service.py
```diff
@@ -0,0 +1,51 @@
...
+    AppServiceCredentials.create(
```
annashamray, line 23:
```suggestion
    AppServiceCredentials.objects.create(
```

src/bptl/work_units/zgw/migrations/0004_auto_20201207_1808.py
```diff
@@ -0,0 +1,17 @@
...
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ("zgw", "0003_auto_20201112_1547"),
+    ]
+
+    operations = [
+        migrations.AlterModelTable(
```
annashamray, line 13: I don't quite understand why there are two `AlterModelTable` migrations with one table?

src/bptl/credentials/forms.py
annashamray: And is it correct that you use Autorisatie API only to get the list of application names? But you don't use credentials since it would not be secure? But in this case why use Autorisatie API at all?

annashamray (changes requested):
I really like that jwt would not be transferred in task variables anymore. It's much neater now.
And thank you for removing all singleton configs and unifying all clients

damm89 (changes requested):
Great refactoring. Much better. :+1:

src/bptl/credentials/forms.py
sergei-maertens: Fixed via f0520669d575de183fa6f62e009e24f6d6d64fbe

sergei-maertens: > And is it correct that you use Autorisatie API only to get the list of application names? But you don't use credentials since it would not be secure? But in this case why use Autorisatie API at all?

Correct. The `app_id` can be any string, but it makes everyone's life easier if we re-use the Autorisaties API URL as ID - as the application can also retrieve that itself from the Autorisaties API so they don't need to exchange that before-hand/make it configurable.

Autorisaties API does not expose credentials of apps, that's a security can of worms I don't want to open.

We use it as convenience - if you can select the app from a pre-defined list, this is less admin overhead.

src/bptl/credentials/forms.py
sergei-maertens: Added a comment pointing to the extra validation.

src/bptl/credentials/forms.py
sergei-maertens: Handled by `return`-ing after field deletion - this code will only kick in when there are applications retrieved from the Autorisaties API and the field is therefore present.

src/bptl/tasks/forms.py
sergei-maertens: I don't really see how - it's a complex mechanism, but it's well tested and contained.

src/bptl/credentials/models.py
```diff
@@ -0,0 +1,143 @@
...
+    # store the required credentials
+    # TODO: use encryption extensions for these fields
+
+    # ZGW client ID + secret
+    client_id = models.CharField(max_length=255, blank=True)
+    secret = models.CharField(max_length=255, blank=True)
+
+    # API key of sorts
+    header_key = models.CharField(_("header key"), max_length=100, blank=True)
+    header_value = models.CharField(_("header value"), max_length=255, blank=True)
```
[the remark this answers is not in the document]
sergei-maertens: The same fields were picked on purpose to keep the similarity.

This is a multi-layered approach. In the case of Open Zaak clients, each has different credentials. However, if we have credentials for BAG or BRP API for example, it's much more likely that we only have a single API Key for the entire organization, and then you don't want to repeat that configuration.

So, what happens is:

* determine services
* check if there are credentials specific to the app (pointed at by `bptlAppId`), use them if they exist
* if there aren't, rely on the Service credentials, which are global to BPTL

You have a multi-staged approach to credentials this way with all the flexibility and the ease of not having to define the same credentials over and over again.

> Perhaps zgw_consumers.Service can be changed: for example, it can be split to base and proxy models with unique constraint only on proxy model (and in this case, it can be backward compatible).

I don't think that adding the extra complexity to zgw_consumers really solves anything - it's just not relevant for other applications using that library. BPTL has a very niche case of managing credentials, and imo the solution for that belongs in BPTL. By default storing the credentials in a separate model in zgw-consumers incurs an INNER_JOIN performance penalty, which is really not needed.

> This case with several credentials for one API can appear somewhere in the future again, I think we need to include its support to zgw_consumers

Let's reconsider if it does appear again in the future.

src/bptl/tasks/forms.py
```diff
@@ -71,6 +70,99 @@ def __init__(self, *args, **kwargs):
...
+        super().clean()
+
+        # selected callback - this is not validated because we don't have access to the
```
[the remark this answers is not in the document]
sergei-maertens: Right - we have a form and a formset, the form is used for the `TaskMapping` model, the formset has the `DefaultService` inline-formset.

From the `DefaultService` formset we can't get to the (validated) `TaskMapping` form instance, they are handled in the view only, but they don't have any references to each other. So, we can't access `form.cleaned_data["callback"]` because we can't get to `form`.

src/bptl/tasks/migrations/0010_auto_20201207_1818.py
```diff
@@ -0,0 +1,17 @@
...
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ("tasks", "0009_auto_20201207_1814"),
+    ]
+
+    operations = [
+        migrations.AlterModelTable(
+            name="defaultservice",
+            table=None,
```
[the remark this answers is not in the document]
sergei-maertens: yeah - this was some magic to move the model from one app to another. `bptl.work_units.zgw.migrations.0004` explicitly sets the table name to `tasks_defaultservice`.

This migration basically un-does that explicit table name specifying, but the end result is the same because the model is in the app `tasks` now, and has the name `defaultservice`.

src/bptl/tasks/migrations/0009_auto_20201207_1814.py
```diff
@@ -0,0 +1,74 @@
...
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ("zgw_consumers", "0011_remove_service_extra"),
+        ("tasks", "0008_auto_20200228_1616"),
+        ("zgw", "0004_auto_20201207_1808"),
+    ]
+
+    operations = [
+        migrations.SeparateDatabaseAndState(
```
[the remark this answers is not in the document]
sergei-maertens: I didn't want to drop/create existing tables.

If you have the `CreateModel` or `RemoveModel` operations, they will create/drop the respective db tables. However, since I am moving the model from one app to another, I just want to rename the DB table and keep the existing data.

So, the migrations do that:

1. rename existing table to table name in new app
2. "remove" the model from the old app
3. "create" the model in the new app

Steps 2. and 3. are only from Django's model perspective, but have no actual changes in the database schema, since we handled that in step 1. already.

src/bptl/work_units/zgw/migrations/0004_auto_20201207_1808.py
sergei-maertens: See https://github.com/GemeenteUtrecht/bptl/pull/96#discussion_r542333268

src/bptl/credentials/models.py
```diff
@@ -0,0 +1,143 @@
...
+    label = models.CharField(
+        _("label"), max_length=100, help_text=_("Human readable app identifier.")
+    )
+    app_id = models.CharField(
+        _("app ID"),
+        max_length=1000,
+        unique=True,
+        help_text=_(
+            "(Globally) unique application identifier. Typically the "
+            "URL of the application in the Autorisaties API."
```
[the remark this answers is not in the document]
sergei-maertens: The API response has a `url` key for each object, which points to the detail endpoint.

src/bptl/work_units/valid_sign/tasks.py
```diff
@@ -107,59 +90,6 @@ class CreateValidSignPackageTask(ValidSignTask):
     * ``packageId``: string. ID of the ValidSign package created by the task.
```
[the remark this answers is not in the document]
sergei-maertens: Yes, need to update that!

src/bptl/work_units/brp/tasks.py
```diff
@@ -1,13 +1,14 @@
...
 from .utils import Relations
 
 __all__ = ["IsAboveAge", "DegreeOfKinship"]
 
 
 @register
+@require_brp_service
 class IsAboveAge(WorkUnit):
     """
     Fetches BRP API and returns whether a person is exactly, or older than, a certain age.
```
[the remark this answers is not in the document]
sergei-maertens: Went over all docstrings and updated them.

src/bptl/work_units/valid_sign/tasks.py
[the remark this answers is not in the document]
sergei-maertens: Went over all docstrings and updated them.

damm89 (approved):
looks good, awesome work!

src/bptl/tasks/forms.py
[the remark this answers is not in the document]
annashamray: Thanks for the explanation

src/bptl/tasks/migrations/0010_auto_20201207_1818.py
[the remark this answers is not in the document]
annashamray: Looks like black magic :)

src/bptl/credentials/models.py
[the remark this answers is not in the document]
annashamray: Ok, I agree that the multi-layered approach fits BPTL case well.

annashamray (approved):

PR #104, opened 2020-12-11
Feature/xential task

### LINKED ISSUE #103: Implement Xential integration
First pass: we use the integration from Contezza.

Webscripts documentation: see https://github.com/GemeenteUtrecht/ZGW/issues/540#issuecomment-726762996
Utrecht webscript index - https://alfresco.utrechtproeftuin.nl/alfresco/s/index/family/xential

Two flows:
- silent
- interactive

Input task variables:
* `nodeRef`
* `templateUuid`
* `filename`
* `templateVariables` (JSON)

Output:
* `buildId` (if available)
* `xentialTemplateUrl` (if available)

annashamray:
Based on https://github.com/GemeenteUtrecht/bptl/pull/96
Fixes #103

### DIFF
--- a/src/bptl/work_units/xential/client.py
+++ b/src/bptl/work_units/xential/client.py
@@ -1,0 +1,30 @@
+"""
+Implements a Xential client.
+"""
+import logging
+
+from django.utils.translation import gettext_lazy as _
+
+from zgw_consumers.constants import APITypes
+
+from bptl.tasks.models import BaseTask
+from bptl.tasks.registry import register
+
+from ..clients import JSONClient, get_client as _get_client
+from ..services import get_alias_service
+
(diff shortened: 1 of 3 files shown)


annashamray:
I'm not sure what to do with Basic auth since it's not supported in `zgw_consumers.Service`, I think the best approach is to include the support to `zgw_consumers` lib

sergei-maertens:
> I'm not sure what to do with Basic auth since it's not supported in zgw_consumers.Service, I think the best approach is to include the support to zgw_consumers lib

Explicit support would be nicer yeah, but you can get around it by using API key with base64 encoded `username:password` value for the header value.

sergei-maertens (approved):

PR #105, opened 2020-12-22
Extract and display Camunda error information

### LINKED ISSUE #81: Extract more error information from Camunda
Camunda often throws HTTP 500 errors if the process execution cannot continue because of mistakes in the process model.

Possibly the response body contains more information than what we currently extract and show. We must see if we can get more info and add it to the BPTL error information.

sergei-maertens:
Fixes #81

* The error response from Camunda is now handled if it's a JSON response
* Error information is saved on the task and displayed in the UI
* Refactored some repetitive markup into sniplates component


src/bptl/dashboard/templates/dashboard/task_detail.html
```diff
@@ -11,103 +11,65 @@
...
 
-            <div class="task-desc__row">
-                <div class="task-desc__name">
-                    Lock expires at
-                </div>
-                <div class="task-desc__value">
-                    {{ task.lock_expires_at }}
-                </div>
-            </div>
+        {% if task|task_type == 'Camunda' %}
```
annashamray, line 49: nice refactoring!

src/bptl/camunda/utils.py
```diff
@@ -149,3 +155,26 @@ def fail_task(task: ExternalTask, reason: str = "") -> None:
...
+        return
+
+    try:
+        error_information = response.json()
+    except json.JSONDecodeError:
+        error_information = None
+
+    TimelineLog.objects.create(
+        content_object=task,
+        template="timeline_logger/camunda/failed_complete.txt",
```
annashamray, line 173: I can't find this template file in the project

BPMN/crash.bpmn
```diff
@@ -0,0 +1,50 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
annashamray, line 1: Why do you need to store this BPMN process here?

annashamray:
Very cool feature!

damm89 (commented):
No more remarks after Anna.

src/bptl/camunda/utils.py
sergei-maertens: it doesn't exist :-) I'll add something

BPMN/crash.bpmn
sergei-maertens: Not really a need - but more so that we all have the same process and don't have to make one ourselves to be able to test/reproduce with.

Issue #110, opened 2021-01-21
SPIKE: investigate Xential + Documenten API integration


sergei-maertens:
## Introduction

[Xential](https://www.xential.com/) provide document creation tools, based on the principle of document templates combined with template variables as input. This templating action can be done in the background as a "silent" template or interactively by an end-user filling out the required fields/variables.

The Documenten API is well-known and part of the standard for API's voor Zaakgericht Werken.

## High level overview of the (processing) flow

Combining all of this in BPTL should lead to the following flow:

1. Provide a task with task variables as input:
    * template ID - which template in Xential to start
    * informatieobjecttype (& possible other required Documenten API fields) - so that we can persist the Xential result document in the Documenten API
    * Xential template variables - to be sent along when starting the Xential template
2. Start the template (silent or interactive)
    * if it's an interactive template, Xential returns a URL for the end-user to interactively create the document
3. Receive the result back from Xential - once the document is created (both silent & interactive). It *looks* like Xential can send back the created document at a particular URL using the webhook principle.
4. Store the result in the documents API
5. (optional) Signal back the process instance via a message event that the document is created (pretty much exactly what the ValidSign implementation does)

## Main goals

1. When you create a ticket in step 2, that ticket is only valid for a limited amount of time. So, template starting/ticket creation should ideally happen just-in-time, when the end-user actually initiates the document creation. Handling this gracefully to reduce friction for end-users is important.
2. It's important that the resulting document from Xential ends up in the Documents API and possibly as a process variable (the URL of said document). Next process steps can then relate the document to the zaak, for example.
3. If this cannot all be implemented/explored in the spike, we need an estimate of the amount of work to achieve this and the feasibility. If it can be implemented roughly (proof-of-concept), an estimate of the remaining amount of work to polish this is required.

## Other

1. The current, Alfresco-based implementation provides a view to list the available Xential templates. We need this to, with a direct connection to Xential. This has the lowest prio and may be considered part of the polishing work.

## Technical hints

* Use the BPTL database and Xential-specific models to store relevant state. E.g. - which process instance correspons with which build ID, mapping to the task so that the variables can be retrieved at a later stage... Look at the ValidSign implementation for inspiration
[...]

PR #111, opened 2021-01-22
#110 - Xential documents

SilviaAmAm:
**Previously** 
Before, in BPTL there was a task that received a nodeRef, a template ID and template variables which it used in a POST request to Alfresco `xential/templates/start` to create a document. Then Alfresco sent the data to Xential and returned back either an ID of a generated document (filling a template silently) or a URL (for a template that needed to be filled by a user).

**Changes**
Now we do the requests directly to Xential API from BPTL. In case of an interactive template, BPTL stores data and returns a BPTL-specific URL so that the 'document starting' process only begins when the user goes to that URL. Once the document is created, Xential sends it to BPTL with a webhook. BPTL then sends the document to the  document API.

*Xential Workflow*:

- Get the `XSessionId` with whoami request
- Create a ticket: all options should be specified (even if they are empty, otherwise you get a 500 error). Template variables for silently building documents also need to be specified in this step (if any are missing, an error occurs). 
The callback URL also needs to be specified here, along with the request that Xential should send when a document is built.

    Silently building case:
    - Start a document: This requires the ticketId. It will return the document uuid (silent document) and a 'resume URL' (interactive document). 
    - Build document: This requires the document uuid. Once the document is built, Xential will notify the webhook.

    Interactively building case:
    - Store the ticket UUID in BPTL. 
    - Return a URL to a BPTL endpoint containing the ticket UUID.
    - When a user navigates to the BPTL URL, the document is started using the ticket UUID. This returns the document uuid (silent document) and a 'resume URL' (interactive document), just as before. In the resume URL, there is a loginTicketUuid parameter, which expires after 15 min (this is not the same UUID as the ticket UUID).
    - Once the document is built interactively, Xential will notify the webhook (just as before).


doc/contrib/xential.rst
```diff
@@ -0,0 +1,45 @@
+Xential
+=======
+
+Integration
+-----------
+
+BPTL can automate document creation using Xential templates.
+
+The `bptl.work_units.xential.tasks.start_xential_template` work-unit needs to know the UUID of the template to
```
sergei-maertens, line 9:
```suggestion
The :func:`bptl.work_units.xential.tasks.start_xential_template` work-unit needs to know the UUID of the template to
```

doc/contrib/xential.rst
```diff
@@ -0,0 +1,45 @@
...
+use and whether to build the document interactively (the user fills the empty fields in the template) or
+'silently'. In the latter case, values to fill the template fields also need to be provided to the work-unit.
```
sergei-maertens, line 11: for silent, they _need_ to be provided indeed, for interactive, they _may_ be provided.

src/bptl/work_units/xential/models.py
```diff
@@ -0,0 +1,24 @@
...
+        _("Xential ticket UUID"), help_text=_("Xential ticket UUID")
+    )
+    task = models.ForeignKey("tasks.BaseTask", on_delete=models.PROTECT)
+
+    class Meta:
+        verbose_name = _("Xential Ticket")
+        verbose_name_plural = _("Xential Tickets")
+
+    def __str__(self):
+        return self.bptl_ticket_uuid
```
sergei-maertens, line 24: Will probably raise an exception since it doesn't return a string, but a `uuid.UUID` instance.

src/bptl/work_units/xential/serializers.py
```diff
@@ -0,0 +1,28 @@
...
+
+from django.utils.translation import gettext_lazy as _
+
+from rest_framework import serializers
+from rest_framework.exceptions import ValidationError
+
+
+class CallbackDataSerializer(serializers.Serializer):
+    bptl_ticket_uuid = serializers.UUIDField(label=_("BPTL ticket UUID"))
+    document = serializers.CharField(help_text=_("Base 64 encoded document content."))
```
sergei-maertens, line 12: Should we use https://github.com/Hipo/drf-extra-fields#base64filefield instead?

src/bptl/work_units/xential/serializers.py
```diff
@@ -0,0 +1,28 @@
...
+        try:
+            # If validate is False, no check is done to see if the data contains non base-64 alphabet characters
+            b64decode(value, validate=True)
+        except binascii.Error as e:
+            if str(e) == "Incorrect padding":
+                raise ValidationError(
+                    _("The provided base64 data has incorrect padding"),
+                    code="incorrect-base64-padding",
+                )
+            raise ValidationError(str(e), code="invalid-base64")
```
sergei-maertens, line 24: I'm afraid if this happens, we'll need to correct the padding on our end rather than rejecting the result.

For now, can you add some logging so that we can see this in sentry if it happens? Not sure if Xential do properly apply the b64 padding or not.

src/bptl/work_units/xential/tasks.py
```diff
@@ -1,28 +1,38 @@
...
 @require_xential_service
 def start_xential_template(task: BaseTask) -> dict:
     """
     Run Xential template with requested variables.
-    The returned value is either buildId of created document (in case it is a silent template)
-    or the url to the page where the user can manually fill in the input fields (in case it is
-    an interactive template).
+    If the ``interactive`` task variable is:
+    * ``"True"``: it returns a URL in ``bptlDocumentUrl`` for building a document interactively
+    * ``"False"``: it returns an empty string in ``bptlDocumentUrl``
```
sergei-maertens, line 24: We should be able to use the native types rather than using strings for True/False: https://docs.camunda.org/manual/latest/user-guide/process-engine/variables/#supported-variable-values

src/bptl/work_units/xential/tasks.py
```diff
@@ -31,43 +41,133 @@ def start_xential_template(task: BaseTask) -> dict:
...
 
-    **Sets the process variables**
+    * ``documentMetadata``: a JSON-object containing the fields required to create a document in the Documenten API.
+        The fields shown below are required.
+
+      .. code-block:: json
+
+         {
+            "bronorganisatie": "string",
+            "creatiedatum": "date",
```
sergei-maertens, line 51: this one can be `date.today()` if not provided

src/bptl/work_units/xential/tasks.py
```diff
@@ -31,43 +41,133 @@ def start_xential_template(task: BaseTask) -> dict:
...
+            "titel": "string",
+            "auteur": "string",
+            "taal": "string",
```
sergei-maertens, line 54: may default to `nld`

src/bptl/work_units/xential/views.py
```diff
@@ -0,0 +1,73 @@
...
+from .handlers import on_document_created
+from .utils import get_xential_base_url, parse_xml
+
+
+class DocumentCreationCallbackView(views.APIView):
+    permission_classes = []
+
+    def post(self, request: Request) -> Response:
+        # The callback sends the base64 encoded document and the BPTL ticket ID as XML.
+        callback_data = parse_xml(request.data)
```
sergei-maertens, line 24: Might be worth to consider https://jpadilla.github.io/django-rest-framework-xml/

src/bptl/work_units/xential/views.py
```diff
@@ -0,0 +1,73 @@
...
+        drc_client = get_client(task, DRC_ALIAS)
+        drc_client.post("enkelvoudiginformatieobjecten", json=document_properties)
+
+        # Notify camunda that the document has been created
+        on_document_created(task)
+
+        return Response(status=status.HTTP_204_NO_CONTENT)
+
+
+class InteractiveDocumentView(View):
```
sergei-maertens, line 49: should inherit from `django.views.generic.RedirectView` and then you only have to implemented `get_redirect_url`

src/bptl/work_units/xential/views.py
```diff
@@ -0,0 +1,73 @@
...
+    def get(self, request: Request, uuid: UUID) -> HttpResponse:
+        # With the BPTL specific UUID, we can retrieve the Xential ticket ID
+        xential_ticket = XentialTicket.objects.get(bptl_ticket_uuid=uuid)
```
sergei-maertens, line 52: you'll want to `get_object_or_404(...)` this query, because I'm the type of person that will be guessing random UUIDs if I know the URL :)

src/bptl/work_units/xential/views.py
```diff
@@ -0,0 +1,73 @@
...
+
+        xential_client = get_client(xential_ticket.task, XENTIAL_ALIAS)
+
+        # Step 1: Retrieve XSessionID
+        xsession_id_url = "auth/whoami"
+        response_data = xential_client.post(xsession_id_url)
+        headers = {"Cookie": f"XSessionID={response_data['XSessionId']}"}
```
sergei-maertens, line 59: I feel this should be part of some `XentialClient` subclass of `JSONClient` to abstract it away in the client itself.

sergei-maertens (changes requested):
Some polishing in this PR and then it can be merged. Feel free to create new issues for larger remarks to tackle them separately.

src/bptl/work_units/xential/tasks.py
SilviaAmAm: I also changed this to be able to use a boolean:
https://github.com/GemeenteUtrecht/bptl/blob/master/src/bptl/tasks/base.py#L13
because otherwise if `variable[name]` is `False`, then we get an error.

src/bptl/work_units/xential/serializers.py
SilviaAmAm: Ok! I have added this because these are the checks that get done by the serializer in open-zaak

src/bptl/work_units/xential/serializers.py
SilviaAmAm: I added this in: https://github.com/GemeenteUtrecht/bptl/pull/126

sergei-maertens (approved):

src/bptl/work_units/xential/views.py
SilviaAmAm: Added this in https://github.com/GemeenteUtrecht/bptl/pull/128

PR #123, opened 2021-02-09
Fixes GemeenteUtrecht/ZGW#628 -- allow empty rol-omschrijving

sergei-maertens:
Allowing the empty variable, and dealing with it by bypassing all
business logic adds more flexibility for process designers.

### DIFF
--- a/src/bptl/work_units/zgw/tasks/rol.py
+++ b/src/bptl/work_units/zgw/tasks/rol.py
@@ -1,4 +1,5 @@
 import logging
+from typing import Optional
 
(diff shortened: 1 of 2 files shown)


annashamray:
Sorry, could you explain why to use `CreateRolTask` if there is no purpose in creating a role?

sergei-maertens:
This is used in loop-constructs in Camunda. See https://github.com/GemeenteUtrecht/ZGW/issues/628

This happens with optional form fields for example.

annashamray (approved):

PR #126, opened 2021-02-10
Improved callback serializer

SilviaAmAm:
The serializer that checks the data sent from Xential now uses the Base64FileField from https://github.com/Hipo/drf-extra-fields#base64filefield

I also added logging for when an error is raised because of incorrect padding.


SilviaAmAm:
Replaced this PR with https://github.com/GemeenteUtrecht/bptl/pull/127

PR #128, opened 2021-02-10
Added XML Parser to Xential callback endpoint

SilviaAmAm:


src/bptl/work_units/xential/views.py
```diff
@@ -5,9 +5,12 @@
 from django.shortcuts import get_object_or_404
 from django.views.generic import RedirectView
 
+from djangorestframework_camel_case.parser import CamelCaseJSONParser
```
sergei-maertens, line 8: unused import?

src/bptl/work_units/xential/views.py
```diff
@@ -5,9 +5,12 @@
...
 from rest_framework import status, views
 from rest_framework.request import Request
 from rest_framework.response import Response
+from rest_framework_xml.parsers import XMLParser
+from rest_framework_xml.renderers import XMLRenderer
```
sergei-maertens, line 13: unused import?

sergei-maertens (changes requested):
As discussed on Slack - the approach with a custom XMLParser that subclasses from `XMLParser`, which then performs the camelCase -> snake_case conversion using the tooling from `djangorestframework_camel_case.utils` should work.

Then you can keep the camelCase in the XML, and deal with snake_case in python.

sergei-maertens (approved):
So clean, it makes me happy

PR #130, opened 2021-02-16
Fixes #114 - Check the status of Xential document builds

### LINKED ISSUE #114: Check for document build errors
After requesting to Xential to build a document, the status of the build can be retrieved through a call to the endpoint:
`/document/{document_id}`. The process should be configured to check for the status. In the case where the response reports that an error has occurred, this should terminate the process. 

To do:

- [x] Create celery beat task that checks the status of the documents in Xential. (similarly to https://github.com/GemeenteUtrecht/bptl/blob/master/src/bptl/conf/base.py#L429)
- [x] Configure task in the settings
- [x] Write tests
- [x] Documentation

SilviaAmAm:
Fixes #114


src/bptl/conf/base.py
```diff
@@ -435,6 +435,10 @@
         # is marked graceful, so it'll just return None and not be scheduled again.
         "schedule": schedule(run_every=10),
     },
+    "task-xential-docs-status-check": {
```
sergei-maertens, line 438:
```suggestion
    "xential-docs-status-check": {
```

src/bptl/work_units/xential/models.py
```diff
@@ -25,6 +25,13 @@ class XentialTicket(models.Model):
         help_text=_("Indicates whether the document was built in Xential."),
         default=False,
     )
+    document_uuid = models.CharField(
```
sergei-maertens, line 28: can't we use a `UUIDField` instead?

src/bptl/work_units/xential/tasks.py
```diff
@@ -175,3 +184,32 @@ def make_xml_from_template_variables(template_variables: dict) -> str:
 
     xml_doc.appendChild(root_element)
     return xml_doc.toxml()
+
+
+@app.task
+def task_check_failed_document_builds():
```
sergei-maertens, line 190:
```suggestion
def check_failed_document_builds():
```

sergei-maertens (changes requested):
Some minor changes

sergei-maertens (approved):

PR #131, opened 2021-02-24
Fixes #116 - Display Xential templates in admin

### LINKED ISSUE #116: Request available Xential templates
There should be some functionality to request the available templates from Xential.

The approach on how to do this is TBC.

SilviaAmAm:
Fixes #116 

Each template group from Xential is now shown as a table in the admin (I used the same template that was used for the Alfresco endpoint).
Each table row shows the name and the uuid of the template.

### DIFF
--- a/src/bptl/work_units/xential/admin_views.py
+++ b/src/bptl/work_units/xential/admin_views.py
@@ -2,7 +2,7 @@ from django.views.generic import TemplateView
 
 from bptl.utils.admin import StaffRequiredMixin
 
-from .client import get_default_clients
+from .client import get_xential_client
 
 
 class TemplatesListView(StaffRequiredMixin, TemplateView):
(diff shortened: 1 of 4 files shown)


src/bptl/work_units/xential/admin_views.py
```diff
@@ -12,13 +12,34 @@ class TemplatesListView(StaffRequiredMixin, TemplateView):
...
-            list_response = client.get(list_url)
-            template_groups[client.api_root] = list_response["data"]["templates"]
+        template_groups = []
+        template_groups_url = "template_utils/getUsableTemplates"
+        xential_client = get_xential_client()
+
+        template_groups_details = xential_client.post(template_groups_url)
+        for template_group in template_groups_details["objects"]:
+            template_group_name = get_name(template_group["fields"])
+            template_group_context = {"name": f"{template_group_name}", "templates": []}
```
sergei-maertens, line 22:
```suggestion
            template_group_context = {"name": template_group_name, "templates": []}
```

src/bptl/work_units/xential/client.py
```diff
@@ -60,11 +61,12 @@ def get_client(task: BaseTask, alias: str) -> "JSONClient":
...
-    clients = [
-        JSONClient(service, service.build_client().auth_header)
-        for service in Service.objects.filter(defaultservice__alias=alias).distinct()
-    ]
-    return clients
+def get_xential_client() -> "XentialClient":
+    default_service = DefaultService.objects.get(alias=XENTIAL_ALIAS)
+    service_credentials = AppServiceCredentials.objects.get(
+        service=default_service.service
+    )
```
sergei-maertens, line 68: There is no guarantee that this object exists - it's possible that the credentials are defined on the `service` only.

Additionally, there's also no guarantee there is only one such object, as different apps may use different credentails for the same service, see the `unique_together` contraint in https://github.com/GemeenteUtrecht/bptl/blob/master/src/bptl/credentials/models.py#L35

`DefaultService` also has no unique constraint on the `alias` field, so you will get `MultipleObjectsReturned`, see https://github.com/GemeenteUtrecht/bptl/blob/master/src/bptl/tasks/models.py#L59

In short - the previous approach where you iterate over all services that have an alias to Xential can all yield valid clients. Probably the approach should be:

* determine all valid services based on the alias
* build clients for each service
* reduce the set of clients to unique clients (i.e. - only keep clients that differ in base endpoint/credentials)

sergei-maertens (changes requested):

sergei-maertens (approved):

PR #154, opened 2021-07-28
:sparkles: Reverse relating of zaken

damm89:


src/bptl/work_units/zgw/tasks/zaak_relations.py
```diff
@@ -288,6 +288,17 @@ def perform(self) -> Optional[dict]:
...
         bijdrage_aard = check_variable(variables, "bijdrageAard")
 
+        # To avoid having to edit BPMN models - default of bijdrage_aard_omgekeerde_richting
+        # is "onderwerp".
+        try:
+            bijdrage_aard_omgekeerde_richting = check_variable(
+                variables, "bijdrageAardOmgekeerdeRichting", empty_allowed=True
+            )
+            if bijdrage_aard_omgekeerde_richting != "":
+                raise ValueError(f"Unknown 'bijdrage_aard': '{bijdrage_aard}'")
```
damm89, line 298:
```suggestion
                raise ValueError(f"Unknown 'bijdrageAardOmgekeerdeRichting': '{bijdrage_aard_omgekeerde_richting}'")
```

src/bptl/work_units/zgw/tasks/zaak_relations.py
```diff
@@ -310,6 +311,44 @@ def perform(self) -> Optional[dict]:
             request_headers=headers,
         )
 
+        try:
+            bijdrage_aard_omgekeerde_richting = check_variable(
+                variables, "bijdrageAardOmgekeerdeRichting", empty_allowed=True
```
annashamray, line 316: Could you add `bijdrageAardOmgekeerdeRichting` to the docstring of the task? (in the "optional process variables" section)

src/bptl/work_units/zgw/tasks/zaak_relations.py
```diff
@@ -310,6 +311,44 @@ def perform(self) -> Optional[dict]:
...
+                )
+        except MissingVariable:
+            # To avoid having to edit BPMN models - default of bijdrage_aard_omgekeerde_richting
+            # is "onderwerp" if it isn't explicitly given in the variables.
+            bijdrage_aard_omgekeerde_richting = AardRelatieChoices.onderwerp
+
+        # Relating of secondary zaken to their main zaak.
+        if (
+            bijdrage_aard != AardRelatieChoices.onderwerp
+            and bijdrage_aard_omgekeerde_richting
```
annashamray, line 333:
```suggestion
            bijdrage_aard != AardRelatieChoices.onderwerp
            
```
`bijdrage_aard_omgekeerde_richting` always has a value (from task variables or the default one)

src/bptl/work_units/zgw/tasks/zaak_relations.py
```diff
@@ -310,6 +311,44 @@ def perform(self) -> Optional[dict]:
...
+                    f"Unknown 'bijdrageAardOmgekeerdeRichting': '{bijdrage_aard_omgekeerde_richting}'"
+                )
+        except MissingVariable:
+            # To avoid having to edit BPMN models - default of bijdrage_aard_omgekeerde_richting
+            # is "onderwerp" if it isn't explicitly given in the variables.
+            bijdrage_aard_omgekeerde_richting = AardRelatieChoices.onderwerp
+
+        # Relating of secondary zaken to their main zaak.
+        if (
+            bijdrage_aard != AardRelatieChoices.onderwerp
```
annashamray, line 332: In the issues there is nothing about the reverse relation should be added only if `aardRelatie` = "onderwerp". It's specified that FB should be able to specify both types of relations. Could you please double check it?

annashamray (changes requested):

annashamray:
@sergei-maertens This is a really interesting case. It looks that perhaps we make a client do the work that should be done by API. Should the reverse relations between cases be created automatically in the API like we do it with `ObjectInformatieObject` and `ZaakBesluit` resources?

src/bptl/work_units/zgw/tasks/zaak_relations.py
damm89: Well, if it is explicitly empty (""), then we don't want to add the reverse relation.

annashamray: oh right!

src/bptl/work_units/zgw/tasks/zaak_relations.py
damm89: I figured it doesn't make sense to add it if the `bijdrage_zaak == "onderwerp"` right? I.e., the main zaak's onderwerp can't be the related zaak if the related zaak's onderwerp is the main zaak. Or am I mistaken?

I guess it should be mutually exclusive, not necessarily only when the `aardrelatie == "onderwerp"`.

annashamray: Afaik there is no check for it in the Zaken API, so technically you can add any combination of `aardrelatie`. Logically I think you are right, but functional beheers should be ok with it, or it would be an unpleasant surprise for them.

sergei-maertens:
@annashamray no - this is not an official pattern from the standard

src/bptl/work_units/zgw/tasks/zaak_relations.py
sergei-maertens: It is logically correct even though technically it's possible. Two cases being the subject of each other does not make sense, so we should avoid getting to that situation.

sergei-maertens (approved):
Please make the change to document the optional variable, after that it's good to merge.

PR #156, opened 2021-07-29
:arrow_up: Bump versions

damm89:


requirements/base.txt
```diff
@@ -1,195 +1,224 @@
...
+babel==2.9.1
+    # via sphinx
+billiard==3.6.4.0
     # via celery
-celery-once==3.0.1
-    # via -r requirements/base.in
-celery==4.4.0
+bitly-api-py3==0.3
+    # via django-docutils
+celery==5.1.2
```
sergei-maertens, line 17: Celery 5 is not supported with Flower, we should stick to Celery 4 for the time being

sergei-maertens (changes requested):

sergei-maertens (approved):

PR #160, opened 2021-09-22
Zgw issue/#908

damm89:
Solves https://github.com/GemeenteUtrecht/ZGW/issues/908


src/bptl/camunda/models.py
```diff
@@ -41,7 +41,7 @@ class ExternalTask(BaseTask):
     )
     priority = models.PositiveIntegerField(_("priority"), null=True, blank=True)
     task_id = models.CharField(_("task id"), max_length=50)
-    instance_id = models.CharField(_("task id"), max_length=50)
+    instance_id = models.CharField(_("process instance id"), max_length=50)
```
sergei-maertens, line 44: please run `makemigrations` for this change as well

src/bptl/camunda/tasks.py
```diff
@@ -79,6 +79,9 @@ def task_execute_and_complete(fetched_task_id):
         logger.warning("Task %r has been already run", fetched_task_id)
         return
 
+    instance_id = fetched_task.instance_id
+    logger.info("Task is part of process instance %d", instance_id)
```
sergei-maertens, line 83:
```suggestion
    logger.info("Task is part of process instance %s", instance_id)
```
-> task instance ID is typically a UUID

src/bptl/dashboard/filters.py
```diff
@@ -19,10 +19,21 @@ class TaskFilter(FilterSet):
...
 
     class Meta:
         model = BaseTask
-        fields = ("status", "topic_name", "engine_type")
+        fields = ("status", "topic_name", "engine_type", "instance_id")
+
+    def filter_by_process_instance_id(self, queryset, name, value: str) -> QuerySet:
+        if not value:
+            return queryset
+        qs = queryset.filter(externaltask__instance_id__iexact=value)
```
sergei-maertens, line 35: maybe allow searching on a part of an ID even?

I also think that you can declare this lookup directly in the filter rather than using `method` argument

```suggestion
        qs = queryset.filter(externaltask__instance_id__icontains=value)
```

sergei-maertens (changes requested):

sergei-maertens (approved):

PR #163, opened 2021-11-17
Skip eigenschap creation if value is empty

sergei-maertens:
This can happen if certain properties are marked as optional.


sergei-maertens:
Frick, just now saw #162

PR #175, opened 2022-03-09
Improve Docker steps in CI

jordi-t:
* Use BuildX/BuildKit actions
* Add caching to speed up image build step (reduced from 4+ minutes to 1+ minute)
* Next to Dockerhub, also push to Harbor


damm89 (approved):
:+1:

PR #176, opened 2022-03-11
:bug: Set  now locks and unlocks the documents

damm89:


src/bptl/work_units/zgw/tasks/documents.py
```diff
@@ -1,5 +1,5 @@
 import logging
-from typing import Dict, List
+from typing import Dict, List, Tuple
```
damm89, line 2:
```suggestion
from typing import Dict, List
```

damm89 (commented):
Remove Tuple import

PR #201, opened 2022-12-15
:bug: Fix edge case bug where 2 zaaktypen exist with an overlapping g…

damm89:
…eldigheid but one has a einde geldigheid more distant into the future


damm89:
Ran tests locally

PR #221, opened 2023-05-03
:sparkles: Let BPTL collect betrokkeneIdentificatie from ZAC in speci…

damm89:
…al cases


damm89:
Fixes https://github.com/GemeenteUtrecht/ZGW/issues/2048
Relies on https://github.com/GemeenteUtrecht/zaakafhandelcomponent/pull/946

